PR #16, opened 2020-06-16
Fix for VTK9 on Mac

### LINKED ISSUE #15: BackgroundPlotter not workin on Mac with vtk9
I *still* cannot get the `BackgrounPlotter` to work with VTK9.0 on Mac. See https://github.com/pyvista/pyvista/issues/562#issuecomment-619428242

cc @Leguark, since you all are using the `BackgroundPlotter` in GemPy, you may want to limit `vtk` in your setup.py until we get this fixed: `vtk<9.0.0`

cc @GuillaumeFavelier, are you all seeing this with MNE-Python?

banesullivan:
Resolve #15 

This works for me locally... someone else should double-check that this is okay (no longer calling initialize at all). Following advice in https://discourse.vtk.org/t/vtk-9-pyqt-macos-no-rendering/3358/14

cc @larsoner and @GuillaumeFavelier

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -380,8 +380,6 @@ class QtInteractor(QVTKRenderWindowInteractorAdapter, BasePlotter):
             istyle = vtk.vtkInteractorStyleTrackballCamera()
             self.SetInteractorStyle(istyle)
 
-            self.iren.Initialize()
-
             self._observers = {}    # Map of events to observers of self.iren
             self._add_observer("KeyPressEvent", self.key_press_event)
 


akaszynski:
Can't test at the moment, but I'll finally be getting a Mac shortly for another project...

GuillaumeFavelier:
I looked at the link and apparently the order is important, `Initialize` needs to be called after `show()`. 

In [blockbuilder](https://github.com/GuillaumeFavelier/blockbuilder/blob/master/blockbuilder/minimal_plotter.py#L22-L25), I created a `showEvent()` function that calls `Initialize` after the event is accepted (not tested on mac).

Another solution would be to create a `show_signal` connected to the `Initialize()` of the interactor.

larsoner:
If we can get away with not calling `Initialize` at all and let VTK figure out the right time to do it (?), then that seems safest. I'll check to see if this fixes things for me on my macOS laptop...

GuillaumeFavelier:
Quoting David Gobbi: "QVTKRenderWindowInteractor.py does not call Start() or Initialize() itself, there’s nothing that can be done there."

Reference: https://discourse.vtk.org/t/vtk-9-pyqt-macos-no-rendering/3358/22

If it's working for everyone without `Initiliaze()`, it's fine by me.

larsoner:
I get a segfault on this branch with the MNE example I usually test with. We could connect to the `show` signal, but at least at the MNE end this will not fix the problem because we still risk creating a bunch of objects before `show` actually gets called.

For now (for us at least) it might be best just to not use VTK9 until they release new wheels with the fix / backport they're doing.

akaszynski:
> For now (for us at least) it might be best just to not use VTK9 until they release new wheels with the fix / backport they're doing.

I'm also having trouble with VTK9 on Linux with the render window refusing to close.
https://gitlab.kitware.com/vtk/vtk/-/issues/17917

Might be best to wait for the next set of wheels.

banesullivan:
> Might be best to wait for the next set of wheels.

Honestly, I have no idea when that could be though...

banesullivan:
@larsoner and @akaszynski, what if we hardset `pyvistaqt` to have `vtk<'9.0.0'` in its `setup.py` until we can fix these things and/or the next set of wheels come out?

larsoner:
It would be good if platform selectors could be used like in `pip` (see for example [this commit](https://github.com/mne-tools/mne-python/pull/7791/commits/e43e9d2a3de87f300de61c5c703ef8f38c0ba915)), but I've never tried to do this in `setup.py` so I'm not sure how feasible it is.

larsoner:
FIX VTK 9.0.1 is now available on PyPi, but it still renders some garbage for me :(

akaszynski:
> FYI VTK 9.0.1 is now available on PyPi, but it still renders some garbage for me :(

Also still having issues with render windows closing on Linux.

banesullivan:
> FYI VTK 9.0.1 is now available on PyPi, but it still renders some garbage for me 

VTK 9.0.1 is working all fine for me on mac with `BackgroundPlotter`. It seems to have fixed all of my issues.. @larsoner, can you share what exactly is going awry for you over in #15

PR #21, opened 2020-06-29
WIP: HiDPI

### LINKED ISSUE #20: BUG: BackgroundPlotter does not handle HiDPI properly
This code:
```
import pyvista as pv
from pyvista import Plotter
# from pyvistaqt import BackgroundPlotter as Plotter

def callback(*args, **kwargs):
    print(args, kwargs)


p = Plotter()
p.add_box_widget(callback)
p.add_mesh(pv.Cone())
p.show()
```
Works just fine for picking and looks like:

Screen Shot 2020-06-29 at 2 41 56 PM

Changing it to use BackgroundPlotter, the window is twice as big (fine), the same number of rendered pixels (I think? if so, bad), and click interactions are off by a factor of 2 (you have to click 50% of the width and 50% of the height to get the right part of the interactor):

Jun-29-2020 14-54-37

There is a lot of code in Mayavi designed to handle this based around `_pixel_ratio`:

https://github.com/enthought/mayavi/blob/master/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py#L250-L253

Presumably some or all of this needs to be adapted to get HiDPI support working. Does this seem like the best way to go?

larsoner:
Closes #20

Not working yet. The RenderWindow.SetSize(...) does not seem to take effect, so resolution is stuck at half on HiDPI screens. In the meantime, forcing `_pixel_ratio=1.` at least fixes the interaction.

@banesullivan do you have a HiDPI screen where you could see if it's possible to get HiDPI rendering to actually work on BackgroundPlotter? I'm not sure what's needed to get that working, but once it does, I think the workarounds here might not be necessary, or uncommenting the `_pixel_ratio = ...` thing might make interactions work properly.

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -15,10 +15,10 @@ import scooby
 from pyvista.plotting.plotting import BasePlotter
 from pyvista.plotting.theme import rcParams
 
-from PyQt5.QtCore import pyqtSignal, pyqtSlot, QObject, QTimer
-from vtk.qt.QVTKRenderWindowInteractor import QVTKRenderWindowInteractor
+from PyQt5.QtCore import pyqtSignal, pyqtSlot, QObject, QTimer, Qt, QEvent
+from tvtk.pyface.ui.qt4.QVTKRenderWindowInteractor import QVTKRenderWindowInteractor
+# from vtk.qt.QVTKRenderWindowInteractor import QVTKRenderWindowInteractor
 from PyQt5 import QtGui
-from PyQt5 import QtCore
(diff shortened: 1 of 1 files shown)


banesullivan:
I'll try this out as soon as I can

akaszynski:
> do you have a HiDPI screen

I've got a 4K screen I can work on.  Once this is passing tests please ping me and I'll take a look.

larsoner:
This probably needs to be fixed at the VTK end. If it's possible at the Python level we can backport it by copying their adapter py code, but it might require C++ changes. We'll see

PR #24, opened 2020-06-30
Documentation updates

banesullivan:


docs/_templates/layout.html
```diff
@@ -0,0 +1,44 @@
...
+{% extends "!layout.html" %}
+
+{% block extrahead %}
+{{ super() }}
+
+<meta name="description" content="3D plotting and mesh analysis through a streamlined interface for the Visualization Toolkit (VTK)">
+<meta name="author" content="PyVista Developers">
+<meta name="keywords" content="python, 3D, visualization, plotting, mesh, meshviewer, vtk, open-source">
+
+<!-- this contains PyVista's Google analytics code - please don't copy/paste our tracking ID -->
```
banesullivan, line 11: Whoops, I need to create a new ID for pyvistaqt

hoechenberger:
Is it intentional that the readme doesn't contain the `conda` installation instruction, while the built documentation does?

banesullivan:
Not intentional - I forgot. Want to include it? On this note, we should maybe update PyVista's as well: https://github.com/pyvista/pyvista#installation

hoechenberger:
> Want to include it?

Currently busy with other stuff, so won't get to do it today!

PR #25, opened 2020-06-30
MAINT: Simpler inheritance

larsoner:
Working on #20 in #21 I started documenting the inheritance and realized that it could probably be simplified by killing `QVTKRenderWindowInteractorAdapter` altogether. This code:
```
import pyvista as pv
from pyvistaqt import BackgroundPlotter as Plotter
from pyvistaqt.plotting import logger
logger.setLevel(10)  # debug


p = Plotter()
p.show()
```
Outputs:
```
BackgroundPlotter init start
QtInteractor init start
BasePlotter init start
BasePlotter init stop
QtInteractor init stop
BackgroundPlotter init stop
```
Which looks like the correct init order to me.

It also:

1. explicitly parents each Qt widget created to what I think should be the correct parent.
2. documents the inheritance and construction order
3. Adds a `self.interactor = self` alias so that code (like MNE) that uses `plotter.interactor.whatever` will still work without needing to be updated.

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -1,35 +1,40 @@
-"""Qt interactive plotter.
+"""
+Diagram
+^^^^^^^
 
-Inheritance
-===========
+.. code-block:: none
 
-BackgroundPlotter
-+-- QtInteractor
-    |-- QVTKRenderWindowInteractor
-    |   +-- QWidget
-    +-- BasePlotter
+    BackgroundPlotter
+    +-- QtInteractor
+        |-- QVTKRenderWindowInteractor
+        |   +-- QWidget
+        +-- BasePlotter
 
-MainWindow
-+-- QMainWindow
+    MainWindow
+    +-- QMainWindow
 
-Usage
-=====
-BackgroundPlotter.__init__(...)
-|-- self.app_window = MainWindow()
-|-- self.frame = QFrame(parent=self.app_window)
-+-- QtInteractor.__init__(parent=self.frame)
-    |-- QVTKRenderWindowInteractor.__init__(parent=parent)
-    |   +-- QWidget.__init__(parent, flags)
-    |-- BasePlotter.__init__(...)
-    +-- self.ren_win = self.GetRenderWindow()
+Implementation
+^^^^^^^^^^^^^^
 
-Because QVTKRenderWindowInteractor calls QWidget.__init__, this will actually
-trigger BasePlotter.__init__ to be called with no arguments. This cannot be
-solved (at least) because using `super()` because
-QVTKRenderWindowInteractor.__init__ does not use super() (also it might not
-be fixable because Qt is doing something in QWidget which is probably separate
-from the super() process). We can fix this by preventing BasePlotter.__init__
-by being called by temporarily monkey-patching it with a no-op __init__.
+.. code-block:: none
+
+    BackgroundPlotter.__init__(...)
+    |-- self.app_window = MainWindow()
+    |-- self.frame = QFrame(parent=self.app_window)
+    +-- QtInteractor.__init__(parent=self.frame)
+        |-- QVTKRenderWindowInteractor.__init__(parent=parent)
+        |   +-- QWidget.__init__(parent, flags)
+        |-- BasePlotter.__init__(...)
+        +-- self.ren_win = self.GetRenderWindow()
+
+Because ``QVTKRenderWindowInteractor`` calls ``QWidget.__init__``, this will
+actually trigger ``BasePlotter.__init__`` to be called with no arguments.
+This cannot be solved (at least) because using ``super()`` because
+``QVTKRenderWindowInteractor.__init__`` does not use ``super()``, and also it
+might not be fixable because Qt is doing something in ``QWidget`` which is
+probably entirely separate from the Python ``super()`` process.
+We fix this by internally by temporarily monkey-patching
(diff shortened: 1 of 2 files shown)


larsoner:
(to get the `BasePlotter init start` to show up, you have to add print statements in `pyvista` as well)

GuillaumeFavelier:
Thanks for your help on this @larsoner, it is much appreciated !

GuillaumeFavelier:
Are you sure `BasePlotter.__init__()` is called only once?

We have a test setup to check that:

https://github.com/pyvista/pyvistaqt/blob/632cbd202134b8d7f48fdbf458c41332d7e7352e/tests/test_plotting.py#L504-L505

When `BasePlotter.__init__()` is called it will add the plotter into a list called `_ALL_PLOTTERS` and if the function is called multiple times, it will add it each time to the list.

This is what I see on Azure ([here](https://dev.azure.com/pyvista/PyVista/_build/results?buildId=3046&view=logs&j=9d81d885-e67f-519e-4101-a242db001def&t=5270f16e-19a4-50a7-5938-3ab8e35bfe4f&l=562) for example):

```
        # check that BasePlotter.__init__() is called only once
>       assert len(_ALL_PLOTTERS) == 1
E       assert 2 == 1
E         +2
E         -1
```

GuillaumeFavelier:
I tested this branch locally and this is what I obtain when I add log in `BasePlotter.__init__()`:

```
BackgroundPlotter init start
QtInteractor init start
BasePlotter init start
BasePlotter init stop
BasePlotter init start
BasePlotter init stop
QtInteractor init stop
BackgroundPlotter init stop
```


This is my config: `python -c "import pyvista; print(pyvista.Report())"`

```
--------------------------------------------------------------------------------
  Date: Wed Jul 01 11:30:14 2020 CEST

                OS : Linux
            CPU(s) : 4
           Machine : x86_64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : GeForce GTX 960M/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 430.64

  Python 3.8.3 (default, May 19 2020, 18:47:26)  [GCC 7.3.0]

           pyvista : 0.25.2
               vtk : 9.0.1
             numpy : 1.18.5
           imageio : 2.8.0
           appdirs : 1.4.4
            scooby : 0.5.5
            meshio : 4.0.13
         pyvistaqt : 0.1.0
             PyQt5 : 5.15.0
           IPython : 7.15.0
             scipy : 1.4.1
              tqdm : 4.46.1
--------------------------------------------------------------------------------
```

larsoner:
Thanks @GuillaumeFavelier, should be fixed now

larsoner:
> LGTM! This a big API change though

FWIW I added the alias `self.interactor = self` so that code (like MNE) that relied on `self.interactor` having specific properties should still work, hopefully minimizing API pain for people. No changes seem to be required in MNE at least with this PR.

pyvistaqt/plotting.py
```diff
@@ -1,9 +1,43 @@
...
+===========
+
+BackgroundPlotter
++-- QtInteractor
+    |-- QVTKRenderWindowInteractor
+    |   +-- QWidget
+    +-- BasePlotter
+
+MainWindow
++-- QMainWindow
```
akaszynski, line 13: Thanks, this was needed.

pyvistaqt/plotting.py
```diff
@@ -339,8 +343,20 @@ def __init__(self, parent=None, title=None, off_screen=None,
...
-        super(QtInteractor, self).__init__(parent=parent, **kwargs)
+        log.debug('QtInteractor init start')
+        # Cannot use super() here because
+        # QVTKRenderWindowInteractor silently swallows all kwargs
+        # because they use **kwargs in their constructor...
+        qvtk_kwargs = dict(parent=parent)
+        for key in ('stereo', 'iren', 'rw', 'wflags'):
+            if key in kwargs:
+                qvtk_kwargs[key] = kwargs.pop(key)
+        with _no_BasePlotter_init():
```
akaszynski, line 354: It's still a hack, but I think it's cleaner than using `QVTKRenderWindowInteractorAdapter` and it's a move in the right direction.

larsoner:
I accepted it but it still says "The base branch restricts merging to authorized users." Maybe I'm not on the right team?

larsoner:
Sorry, one more change. Now that QtInteractor inherits from (eventually) `QWidget`, it should not set `self.parent = parent`. `parent()` is a method of QWidget.

akaszynski:
> I accepted it but it still says "The base branch restricts merging to authorized users." Maybe I'm not on the right team?

Weird, because @GuillaumeFavelier is also part of developers and can also push branches.  Maybe the branch has to be part of the `pyvista` repo?

Can you please make a dummy pr (a comment to `__init__.py`) so I can get to the bottom of this?

pyvistaqt/plotting.py
banesullivan: It'd be good to autodoc this in the docs. @larsoner, could you pull the changes on master and add this somewhere in the API reference page with an autodoc

pyvistaqt/plotting.py
banesullivan: This is a solid case for monkey patching

larsoner:
Added to docs, I'll merge once everything comes back happy

PR #26, opened 2020-06-30
ENH: Add pinch-to-zoom

### LINKED ISSUE #19: ENH: Pinch-to-zoom support
We want to implement pinch-to-zoom support in MNE-Python. Would it make sense to add it directly in pyvistaqt in `QtInteractor` or `BackgroundPlotter`? From what I (incompletely) understand so far it will probably require something like:
```
    def __init__(self, ...):  # init of the the window, might work at the widget level, too; TBD
        ...
        self.grabGesture(Qt.PinchGesture)
        ...

    def event(self, event):  # override general event handler
        """Handle events."""
        if event.type() == QtCore.QEvent.Gesture:
            return self.gestureEvent(event)
        return super().event(event)

    def gestureEvent(self, event):
        """Handle gesture events."""
        ...  # actually do the heavy lifting to handle start, during, and stop stages of the gesture
```

larsoner:
Closes #19

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -405,6 +405,13 @@ class QtInteractor(QVTKRenderWindowInteractorAdapter, BasePlotter):
         self._first_time = False # Crucial!
         self.view_isometric()
 
+    def gesture_event(self, event):
+        pinch = event.gesture(QtCore.Qt.PinchGesture)
+        if pinch:
+            self.camera.Zoom(pinch.scaleFactor())
+            event.accept()
+        return True
+
     def key_press_event(self, obj, event):
         """Call `key_press_event` using a signal."""
         self.key_press_event_signal.emit(obj, event)
@@ -669,6 +676,8 @@ class BackgroundPlotter(QtInteractor):
         super(BackgroundPlotter, self).__init__(parent=self.frame,
                                                 off_screen=off_screen,
                                                 **kwargs)
+        self.app_window.grabGesture(QtCore.Qt.PinchGesture)
+        self.app_window.signal_gesture.connect(self.gesture_event)
         self.app_window.signal_close.connect(self._close)
 
         self.main_menu = None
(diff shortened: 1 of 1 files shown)


akaszynski:
@larsoner, once you've been approved as a developer, you should be able to merge the PR.  Let me know if you have any issues.

larsoner:
@hoechenberger can you test and verify that it works for you ?

hoechenberger:
@larsoner 
> @hoechenberger can you test and verify that it works for you ?

Will give it a shot in a couple hours and report back

hoechenberger:
It's working 😍🥳🥂 Could be a little bit smoother (smaller steps, perhaps?) but otherwise works as expected!! 

Screen-Recording-2020-07-02-at-13 40 49

I suppose continuous update like shown below (with reduced quality until the action is finished) wouldn't be possible..?

Untitled

larsoner:
I think that's just as fast as the events get processed by qt. I experience the same thing with the scroll wheel

banesullivan:
Smooth scrolling is possible while holding the control key (mac) plus click and drag. You may be able to use those same events to control the smoothness?

hoechenberger:
> Smooth scrolling is possible while holding the control key (mac) plus click and drag. You may be able to use those same events to control the smoothness?

Indeed this does work! I keep pressing Ctrl and then use pinch-to-zoom, and it's smooth. I think this behavior should be the default for pinch zooming :)

banesullivan:
I'm not sure if we can (or should) change that default zoom style. There may be performance concerns. The change would likely be upstream in PyVista though

hoechenberger:
> I'm not sure if we can (or should) change that default zoom style. There may be performance concerns. The change would likely be upstream in PyVista though

I'm just talking about the new pinch zooming :) All other existing zoom functionality can remain the same.

hoechenberger:
> Smooth scrolling is possible while holding the control key (mac) plus click and drag.

I can also get this behavior by pressing right mouse button and then scrolling.

banesullivan:
You may be able to trigger the control key events with what was implemented here to do it

PR #36, opened 2020-07-16
Add Makefile targets for coverage

GuillaumeFavelier:
This small PR brings some conveniance targets to the `Makefile`.

It's just so much easier to type `make coverage` instead of remembering the full command everytime.


GuillaumeFavelier:
I plan to work on coverage after this PR so this is a good addition to the workflow for me :)

GuillaumeFavelier:
This is now ready for reviews @pyvista/developers !

PR #37, opened 2020-07-16
Improve coverage

GuillaumeFavelier:
This PR improves the coverage. I'll see how far I can go.

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -273,8 +273,7 @@ class ScaleAxesDialog(QDialog):
 def resample_image(arr, max_size=400):
     """Resample a square image to an image of max_size."""
     dim = np.max(arr.shape[0:2])
-    if dim < max_size:
-        max_size = dim
+    max_size = min(max_size, dim)
     x, y, _ = arr.shape
     sx = int(np.ceil(x / max_size))
     sy = int(np.ceil(y / max_size))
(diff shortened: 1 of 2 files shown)


GuillaumeFavelier:
I'm surprised there is no need for synchronization to test depth peeling and offscreen rendering. I would expect `waitExposed` to be needed, especially on MacOS. Good if it works without it.

Next is smoothing (polygon, line, etc...). From my experience this one is not compatible with Azure. I'll try anyway.

Except from that, the only part remaining for testing is drag & drop plus pinch gesture. AFAIK, those events are not available in `pytest-qt` so a hack will be necessary here.

larsoner:
> Except from that, the only part remaining for testing is drag & drop plus pinch gesture. AFAIK, those events are not available in pytest-qt so a hack will be necessary here.

It would be better to first look into possibly adding them to `pytest-qt`. Not sure how complex it would be, though.

GuillaumeFavelier:
> It would be better to first look into possibly adding them to pytest-qt

Sounds like a plan.

That will be all for this PR then.

More comments:
- I do not add any mesh in `test_depth_peeling`, `test_off_screen` and `test_smoothing`.
- Many tests use `test_background_plotting` in their name, I don't think it's necessary anymore.

GuillaumeFavelier:
This is now ready for reviews @pyvista/developers !

larsoner:
Thanks @GuillaumeFavelier

Issue #30, opened 2020-07-03
Code style


GuillaumeFavelier:
I think the [same question](https://github.com/pyvista/pyvista/issues/288) can be asked in this repo. And eventually the same solution could be applied.

banesullivan:
It'd be easy to do it on this repo. I've used Black on other projects and have been enjoying it

tkoyama010:
+1 for using Black.

GuillaumeFavelier:
I notice some *real motivation* behind `black` 😅

To be honest, my nvim flake8 plugin is crying when I develop for pyvista and pyvistaqt so at this point, I am in favour of anything that can solve this 🙏

So, how does it work in terms of integration? What would be the plan?

tkoyama010:
I made a PR which add CI of checking by black. Could you give opinion?

GuillaumeFavelier:
Closing by #38

PR #38, opened 2020-07-17
:new: black CI by GitHub Actions #30

tkoyama010:
:new: black CI by GitHub Actions for issue #30
In this PR. We need to run GitHub Actions.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -15,13 +15,9 @@ jobs:
       uses: ricardochaves/python-lint@v1.1.0
       with:
         python-root-list: pyvistaqt
-        use-pylint: true
-        use-pycodestyle: true
-        use-flake8: true
+        use-pylint: false
+        use-pycodestyle: false
+        use-flake8: false
         use-black: true
-        use-mypy: true
-        use-isort: true
-        extra-pylint-options: --disable=F0401
-        extra-pycodestyle-options: --max-line-length=88 --ignore="W503"
-        extra-mypy-options: --no-implicit-optional
-        extra-flake8-options: --max-line-length=88
+        use-mypy: false
+        use-isort: false
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -1,4 +1,4 @@
-# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# This workflow will run lint
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 name: Python package
 on:
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -1,0 +1,27 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+name: Python package
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Python Code Quality and Lint
+      uses: ricardochaves/python-lint@v1.1.0
+      with:
+        python-root-list: meshioplt
+        use-pylint: false
+        use-pycodestyle: false
+        use-flake8: false
+        use-black: true
+        use-mypy: false
+        use-isort: false
+        extra-pylint-options: --disable=F0401
+        extra-pycodestyle-options: --max-line-length=88 --ignore="W503"
+        extra-mypy-options: --no-implicit-optional
+        extra-flake8-options: --max-line-length=88
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Python Code Quality and Lint
       uses: ricardochaves/python-lint@v1.1.0
       with:
-        python-root-list: meshioplt
+        python-root-list: pyvistaqt
         use-pylint: false
         use-pycodestyle: false
         use-flake8: false
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -15,9 +15,9 @@ jobs:
       uses: ricardochaves/python-lint@v1.1.0
       with:
         python-root-list: pyvistaqt
-        use-pylint: false
-        use-pycodestyle: false
-        use-flake8: false
+        use-pylint: true
+        use-pycodestyle: true
+        use-flake8: true
         use-black: true
         use-mypy: false
         use-isort: false
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -19,8 +19,8 @@ jobs:
         use-pycodestyle: true
         use-flake8: true
         use-black: true
-        use-mypy: false
-        use-isort: false
+        use-mypy: true
+        use-isort: true
         extra-pylint-options: --disable=F0401
         extra-pycodestyle-options: --max-line-length=88 --ignore="W503"
         extra-mypy-options: --no-implicit-optional
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ CODESPELL_IGNORE ?= "ignore_words.txt"
 
 all: doctest
 
-doctest: black codespell pydocstyle
+doctest: codespell pydocstyle
 
 black:
 	@echo "Running black"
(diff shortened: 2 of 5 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,27 @@
...
+    branches: [ master ]
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Python Code Quality and Lint
+      uses: ricardochaves/python-lint@v1.1.0
+      with:
+        python-root-list: meshioplt
```
tkoyama010, line 17:
```suggestion
        python-root-list: pyvistaqt
```

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,27 @@
...
+        use-pylint: false
+        use-pycodestyle: false
+        use-flake8: false
+        use-black: true
+        use-mypy: false
+        use-isort: false
+        extra-pylint-options: --disable=F0401
+        extra-pycodestyle-options: --max-line-length=88 --ignore="W503"
+        extra-mypy-options: --no-implicit-optional
+        extra-flake8-options: --max-line-length=88
```
GuillaumeFavelier, line 27: Since `pylint`, `flake8`, `mypy` and `pycodestyle` are disabled, I don't think the extra options are necessary. Overall, I would just keep it with the defaults:

```py
      with:
        python-root-list: pyvistaqt
        use-black: true
```

GuillaumeFavelier: Looking at https://github.com/ricardochaves/python-lint/blob/master/action.yml, all of that is enabled by default so you're right, the following seems necessary:

```py
        use-pylint: false
        use-pycodestyle: false
        use-flake8: false
        use-mypy: false
        use-isort: false
```

tkoyama010: Thanks for your comment. I think also pylint, pycodestyle, flake8, mypy, isort is also useful. How about set enable these option too?

tkoyama010:
@GuillaumeFavelier That is a good question. I don't know it either :thinking: I plan to do some research :mag:

tkoyama010:
My guess is that this branch is in my fork repository. In other words, you could create a test branch in pyvistaqt and check it there. If you could make a branch, I will reconfigure this PR to merge into it.

tkoyama010:
This branch is passed in my for repository :+1:

GuillaumeFavelier:
Could you explain what happened? I can see that the github action passed on your `patch-1` branch in your repo but I thought that it would be triggered only on `master`.

Does it mean that there will be a commit "reformat by black" on each PR from now on?

Sorry for the noise, I'm just not used to `black`.

tkoyama010:
@GuillaumeFavelier Sorry, my explanation was not enough. This CI is just to check if the PR source code follows black. I pushed the new commit "reformat by black" by myself to fix the CI error :bowing_man:

.github/workflows/pythonpackage.yml
larsoner: This comment has not been resolved, @GuillaumeFavelier what do you prefer?

I agree they should be enabled, or their options removed.

larsoner:
Otherwise looks fine to me

.github/workflows/pythonpackage.yml
tkoyama010: @larsoner Thanks I will set it enabled.

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Python Code Quality and Lint
+      uses: ricardochaves/python-lint@v1.1.0
+      with:
+        python-root-list: pyvistaqt
+        use-pylint: false
+        use-pycodestyle: false
+        use-flake8: false
```
tkoyama010, line 20:
```suggestion
        use-pylint: true
        use-pycodestyle: true
        use-flake8: true
```

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,27 @@
...
+        use-black: true
+        use-mypy: false
+        use-isort: false
```
tkoyama010, line 23:
```suggestion
        use-mypy: true
        use-isort: true
```

tkoyama010:
Please wait while CI pass in https://github.com/tkoyama010/pyvistaqt/pull/1 .

tkoyama010:
If we use all enable option we have to solve this CI error.
https://github.com/tkoyama010/pyvistaqt/pull/1/checks?check_run_id=881454932

.github/workflows/pythonpackage.yml
tkoyama010: I think this time. It it good to add only black option and remove extra option.

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,27 @@
...
+        use-pylint: true
+        use-pycodestyle: true
+        use-flake8: true
+        use-black: true
+        use-mypy: true
+        use-isort: true
+        extra-pylint-options: --disable=F0401
+        extra-pycodestyle-options: --max-line-length=88 --ignore="W503"
+        extra-mypy-options: --no-implicit-optional
+        extra-flake8-options: --max-line-length=88
```
tkoyama010, line 27:
```suggestion
        use-pylint: false
        use-pycodestyle: false
        use-flake8: false
        use-black: true
        use-mypy: false
        use-isort: false
```

GuillaumeFavelier:
We can still make `pylint` happy in a follow up PR if necessary :+1:

.github/workflows/pythonpackage.yml
tkoyama010: Enable all option has a lot of error. I think it is not a good idea to enable all option. I will remove extra option.

tkoyama010:
> We can still make `pylint` happy in a follow up PR if necessary 👍
OK. Let's make pylint PR next time. Could you merge this PR first?

GuillaumeFavelier:
> Could you merge this PR first?

Sure, I would just like to wait a little for @akaszynski or @banesullivan opinion before merging if you don't mind.

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,23 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
```
banesullivan, line 1: Just to be clear, this doesn't run any of the actual tests, right?

akaszynski:
I'm also an favor of this.  We all have a coding style and it would be great to have a truly uniform one across the project.

I'm going to miss the single quotes though...

akaszynski:
Would also like a makefile that runs black.  @tkoyama010, think you could add that?

banesullivan:
> I'm going to miss the single quotes though...

Same. `'` >> `"`. But with Black, it doesn't even matter. Code how you like, run Black and it should fix em. We can also switch it to prefer single quotes, but they have [pretty good reasoning for double quotes](https://black.readthedocs.io/en/stable/the_black_code_style.html#strings)

> Would also like a makefile that runs black.

[Tox](https://tox.readthedocs.io/en/latest/) is also another way of doing this, but @tkoyama010, do whatever is most simple

akaszynski:
> Why settle on double quotes? They anticipate apostrophes in English text. They match the docstring standard described in PEP 257. An empty string in double quotes ("") is impossible to confuse with a one double-quote regardless of fonts and syntax highlighting used. On top of this, double quotes for strings are consistent with C which Python interacts a lot with.

> On certain keyboard layouts like US English, typing single quotes is a bit easier than double quotes. The latter requires use of the Shift key. My recommendation here is to keep using whatever is faster to type and let Black handle the transformation.

As I'm working with C more on my "real job", it's a bit annoying to switch between the two, so I'll agree that there's value with that.

Tox is indeed another option as well.  I like how it installs the testing env, etc.

.github/workflows/pythonpackage.yml
tkoyama010: You are right.
```suggestion
# This workflow will run lint
```

tkoyama010:
> @tkoyama010, can you add a `makefile` (or `tox`) to do this testing locally?

+1 I'll add this.

tkoyama010:
Does anyone know what the CI error means?

tkoyama010:
The CI error is an HTTP error, so it seems to be resolved by retrying.

tkoyama010:
@pyvista/developers It is ready for merge :+1:

akaszynski:
@banesullivan, I think this one needs your approval.  I'm happy with this.

tkoyama010:
@GuillaumeFavelier @larsoner @banesullivan @akaszynski  Thank you. Now that we have reached an agreement, could you please merge it?

PR #47, opened 2020-07-31
💚 Add mypy in CI

tkoyama010:
Let's add [mypy](https://github.com/python/mypy) in CI. We need type hint for this.
mypy logo


GuillaumeFavelier:
As described in [mypy documentation](https://mypy.readthedocs.io/en/stable/getting_started.html#function-signatures-and-dynamic-vs-static-typing), by default `mypy` will assume the functions to be dynamically typed and only check the imported modules. I went ahead and ignored those imports in favor of `--disallow-untyped-defs`. I think it suits better your intention @tkoyama010 ?

tkoyama010:
Thanks for @GuillaumeFavelier 's work.

pyvistaqt/dialog.py
```diff
@@ -78,7 +84,7 @@ class DoubleSlider(QSlider):
 
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: None, **kwargs: None) -> None:
```
tkoyama010, line 87:
```suggestion
    def __init__(self, *args: Any, **kwargs: Any) -> None:
```

pyvistaqt/plotting.py
```diff
@@ -265,29 +269,29 @@ def __init__(
         self.view_isometric()
         LOG.debug("QtInteractor init stop")
 
-    def gesture_event(self, event):
+    def gesture_event(self, event: Any) -> bool:
         """Handle gesture events."""
```
tkoyama010, line 273:
```suggestion
        """Handle gesture events."""
```

pyvistaqt/plotting.py
```diff
@@ -265,29 +269,29 @@ def __init__(
         self.view_isometric()
         LOG.debug("QtInteractor init stop")
 
-    def gesture_event(self, event):
+    def gesture_event(self, event: Any) -> bool:
```
tkoyama010, line 272:
```suggestion
    def gesture_event(self, event: QGestureEvent) -> bool:
```

tkoyama010:
:point_up: This error is the Type Annotation problem of Python3.5. I think we should finish support of Python3.5.

pyvistaqt/plotting.py
```diff
@@ -729,64 +750,69 @@ def set_icon(self, img):
...
+    def __del__(self) -> None:  # pragma: no cover
         """Delete the qt plotter."""
         if not self._closed:
             self.app_window.close()
 
-    def add_callback(self, func, interval=1000, count=None):
+    def add_callback(
+        self,
+        func: Callable,
+        interval: Optional[int] = 1000,
```
tkoyama010, line 804:
```suggestion
        interval: int = 1000,
```

tkoyama010:

akaszynski:
I'm all for static type checking, especially since it's optional.  While this won't replace unit testing, it can help us identify errors (and help the IDE).  I'd like to hear the opinions of the other contributors as well.

pyvistaqt/plotting.py
```diff
@@ -161,28 +170,28 @@ class QtInteractor(QVTKRenderWindowInteractor, BasePlotter):
...
+        parent: MainWindow = None,
+        title: str = None,
+        off_screen: bool = None,
+        multi_samples: int = None,
+        line_smoothing: bool = False,
+        point_smoothing: bool = False,
+        polygon_smoothing: bool = False,
+        auto_update: Union[float, bool] = 5.0,
+        **kwargs: Any
+    ) -> None:
```
akaszynski, line 185: It's great that `sphinx` can build the docs from this and we don't have to include the type definitions within the doc string!

tkoyama010: @akaszynski That is the cool work of @tk0miya .

banesullivan (commented):
I haven't always been a fan of type annotations but lately, they have been growing on me. Thanks for taking this on to move pyvistaqt in a good direction, @tkoyama010! 

This works well here because it is a smaller codebase, but I think it would be tough to do in pyvista core

akaszynski:
> This works well here because it is a smaller codebase, but I think it would be tough to do in pyvista core

Agreed.  We can just do it in chunks since `mypy` is optional.

akaszynski (approved):

pyvistaqt/plotting.py
```diff
@@ -302,7 +314,8 @@ def dragEnterEvent(self, event):  # pylint: disable=invalid-name,no-self-use
         except IOError as exception:  # pragma: no cover
             warnings.warn("Exception when dropping files: %s" % str(exception))
 
-    def dropEvent(self, event):  # pylint: disable=invalid-name
+    # pylint: disable=invalid-name,useless-return
+    def dropEvent(self, event: QtCore.QEvent) -> None:
```
larsoner, line 318: Why bother adding the `-> None`? Can it be omittted and the `return None` removed / not added? Or is it required by mypy to always have the return values even when None?

tkoyama010: Thank you for comment. As you say mypy always have the return values even when None.

larsoner (approved):

Issue #58, opened 2020-09-22
🆙 PyQt5 version up.


tkoyama010:
:up: Let's split PyQt5 version up Pull Request.

_Originally posted by @tkoyama010 in https://github.com/pyvista/pyvistaqt/pull/51#issuecomment-696480605_

larsoner:
I think we have multiple versions in CI now so I'll close

PR #59, opened 2020-09-26
📦 PyQt5==5.15.1

tkoyama010:
See #58

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -6,7 +6,7 @@ pytest
 pytest-cov
 pytest-memprof
 codecov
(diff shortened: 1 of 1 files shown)


GuillaumeFavelier:
As I said in https://github.com/pyvista/pyvistaqt/pull/81#issuecomment-776727361, I tried updating the CIs operating system to no avail. There is something going on those Linux conf on Azure. I work on Linux regularly with the latest version of PyQt5 and everything works fine, so there should be a way.

larsoner:
We now test a 5.15 variant I think

PR #12, opened 2020-06-05
🚧 Add basic qtpy layer

GuillaumeFavelier:
I think this PR could help https://github.com/pyvista/pyvistaqt/issues/8 by using a thin qt abstraction layer.

The goal would be to have one code and change the `QT_API` accordingly or possibly auto-detect it.
Basic testing with `PySide2` would be nice to have too.

PR #61, opened 2020-10-07
Feat/qtpy

nicobako:
This is a continuation of @GuillaumeFavelier 's work from #12 . I have updated his work with the latest _master_ branch from pyvista, which had some merge conflicts and required replacing `pyqt5` with `qtpy` in a few additional places.

I also updated the test files to use `qtpy` instead of `pyqt5`. I'm not exactly sure how your CI works, but I imagine that you will have to set the `QT_API` environment variable to inform _qtpy_ which Qt Binding you want to use. I'm not sure if you want your CI to test multiple QT bindings, or just one... I can definitely work on this if you let me know which Qt Bindings you want to test.

Also, I replaced the `pyqt5` requirement in the `install_requires` with `QtPy`. This makes the user responsible for installing the Qt bindings of their choice, and makes _pyvistaqt_ agnostic of the specific Qt bindings.

Another thing that we must still do is update the documentation by replacing any usage of `pyqt5` with `qtpy`. I wasn't exactly sure how you wanted to treat that, so I haven't done it yet.
Users will have to use `qtpy` instead, which has the additional step of setting the `QT_API` environment variable. From what I understand, this must be set before _qtpy_ is imported. I can easily update the docs if you let me know what kind of guidance you want to give to users.

So, in summary, things that still need to be done are:

- [ ] Update tests to set `QT_API` environment variable
    - [ ] Should probably test multiple Qt Bindings (e.g. pyqt5, PySide2, etc.)
- [ ] Update docs to use `qtpy` instead of `pyqt5`
    - [ ] Update examples in `usage.rst`
    - [ ] Maybe update example in `index.rst`?
    - [ ] Update license info in `index.rst`


pyvistaqt/counter.py
```diff
@@ -1,7 +1,4 @@
-"""
-This module contains a basic Qt-compatible counter class.
-"""
```
GuillaumeFavelier, line 3: The docstring has been removed in the process

pyvistaqt/counter.py
```diff
@@ -25,8 +22,8 @@ def __init__(self, count: int) -> None:
         else:
             raise ValueError("count is not strictly positive.")
 
-    @pyqtSlot()
-    def decrease(self) -> None:
```
GuillaumeFavelier, line 29: The type hint should remain too

pyvistaqt/plotting.py
```diff
@@ -49,20 +49,20 @@
 import pyvista
 import scooby  # type: ignore
 import vtk
-from PyQt5 import QtCore
-from PyQt5.QtCore import QSize, QTimer, pyqtSignal
-from PyQt5.QtWidgets import (  # pylint: disable=unused-import
+from pyvista.plotting.plotting import BasePlotter
+from pyvista.plotting.theme import rcParams
+from pyvista.utilities import conditional_decorator, threaded
+from qtpy import QtCore, QtGui
```
GuillaumeFavelier, line 55: I think importing `QtGui` is not necessary here

GuillaumeFavelier (commented):
I'm not sure of the issue with MacOS configs on Azure though. It seems related to the changes to `setup.py`.

nicobako:
Yeah, @GuillaumeFavelier. This can easily be fixed by ensuring that a Qt Python binding is installed on the test machine, but I'm not sure exactly how you want to do that. Switching to qtpy  means that it is now up to the _user_ to have a qtpy-compatible Qt Python binding installed. In my opinion, it no longer makes sense to add a specific Qt Python binding in the _install requires_, since that is now up to the user.

GuillaumeFavelier:
Historically, `PyQt5` has been the default Qt binding for `pyvistaqt` and more importantly, this is the only Qt binding tested at the moment. I would not consider removing it until at least one other Qt binding is tested. So I don't think we are ready yet to entirely change this.

In my opinon, a "transition time" (one major version?) where `PyQt5` would remains with `QtPy` is necessary instead of a drastic change.

Does it make sense?

Also, what are your thoughts about this @pyvista/developers ?

pyvistaqt/dialog.py
```diff
@@ -27,9 +27,7 @@ class FileDialog(QFileDialog):
     the dialog was property closed.
     """
 
-    # pylint: disable=too-few-public-methods
```
GuillaumeFavelier, line 30: Restoring this should make the Github Actions CI finally happy :)

nicobako:
> In my opinon, a "transition time" (one major version?) where `PyQt5` would remains with `QtPy` is necessary instead of a drastic change.
> 
> Does it make sense?

This makes sense to me. That way, when users `pip install pyvistaqt`, they still get the same functionality they are used to, but can start experimenting with different Qt bindings on their own.

pyvistaqt/dialog.py
nicobako: I'm on it!

nicobako:
@GuillaumeFavelier , you'll see that I made quite a few changes since your last review. Most notably, updated the docs, and formatting issues, fixed the call to pydoctest in the makefile.

The only issue left is that the docs are still not building. I'm not sure why, but maybe you understand the error message...

GuillaumeFavelier:
> you'll see that I made quite a few changes since your last review. Most notably, updated the docs, and formatting issues, fixed the call to pydoctest in the makefile.

Those changes are welcome, thank you @nicobako :pray: 

> The only issue left is that the docs are still not building. I'm not sure why, but maybe you understand the error message...

I'll see what I can do

GuillaumeFavelier:
I'm sure I saw this error before somewhere but I can't recall exactly...

```
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.

Fatal Python error: Aborted

Current thread 0x00007faf5ef7e700 (most recent call first):
  File "/home/vsts/work/1/s/pyvistaqt/plotting.py", line 600 in __init__
  File "<doctest default[2]>", line 1 in <module>
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/doctest.py", line 1337 in __run
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/doctest.py", line 1483 in run
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/ext/doctest.py", line 554 in test_group
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/ext/doctest.py", line 470 in test_doc
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/ext/doctest.py", line 366 in write
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 361 in build
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 299 in build_update
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/application.py", line 348 in build
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/cmd/build.py", line 280 in build_main
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/cmd/make_mode.py", line 155 in run_generic_build
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/cmd/make_mode.py", line 167 in run_make_mode
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/cmd/build.py", line 199 in make_main
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/cmd/build.py", line 292 in main
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/__main__.py", line 15 in <module>
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/runpy.py", line 85 in _run_code
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/runpy.py", line 193 in _run_module_as_main
Aborted (core dumped)
```

Is it related to the headless display? Any idea @pyvista/developers 
I would rather not hack something while being in the dark :sweat_smile:

larsoner:
There are a slew of Linux deps that you need to use the builtin xcb (PyQt5 used to bake this into their lib IIRC then they changed), for example see what we install on CircleCI for MNE

https://github.com/mne-tools/mne-python/blob/master/.circleci/config.yml#L58-L59

IIRC that big long line is what's needed for this not to complain. These were identified by doing `LD_DEBUG=libs ...` and then seeing what wasn't found while on an interactive SSH CircleCI session (though in theory you could do the same locally just by pulling their docker image).

GuillaumeFavelier:
Many thanks for the swift answer! I will try with that then :+1:

GuillaumeFavelier:
Alright @nicobako, I pushed a few commits and the doc is building now :+1: 
Let us know when it's okay on your end.

Useful material for future reference: https://doc.qt.io/qt-5/linux-requirements.html

nicobako:
Thanks @GuillaumeFavelier . I think it's all good on my end. As long as you guys are happy with the changes I made to the documentation (*.rst* files), then I think this is ready!

GuillaumeFavelier (approved):
LGTM and personally, I like the changes to the *.rst files but let's ask the other @pyvista/developers to be sure.

docs/index.rst
```diff
@@ -84,6 +86,16 @@ sphere.
...
+``pyvistaqt`` is under the MIT license.
+However, Qt bindings have licenses of their own.
+
+Historically, ``pyvistaqt`` has used ``pyqt5``, which is subject
+to the GPL license. See details at
 `Riverbank License FAQ <https://www.riverbankcomputing.com/commercial/license-faq>`_.
+
+``pyvistaqt`` is transitioning to using ``qtpy``
+
+> QtPy is a small abstraction layer that lets you write applications using a single API call to either PyQt or PySide.
```
larsoner, line 98: Why the leading `>`? I would make this more compact like:
```suggestion
Historically, ``pyvistaqt`` has used ``pyqt5``, which is subject
to the GPL license. See details at
`Riverbank License FAQ <https://www.riverbankcomputing.com/commercial/license-faq>`_. ``pyvistaqt`` now uses QtPy if present, and will require it after version 0.3. QtPy is a small abstraction layer that lets you write applications using a single API call to either PyQt or PySide (which has a LGPL license).
```
The wording about being required or not should be adjusted based on whether or not there is actually going to be a deprecation period.

pyvistaqt/counter.py
```diff
@@ -1,17 +1,14 @@
-"""
-This module contains a basic Qt-compatible counter class.
-"""
-from PyQt5.QtCore import QObject, pyqtSignal, pyqtSlot
+"""This module contains a basic Qt-compatible counter class."""
+
+from qtpy.QtCore import QObject, Signal, Slot
```
larsoner, line 3: This will immediately break any `pyvistaqt` code without deprecation. We colud make a small compatibility layer just by making a little `._qtpy` file that contains:
```
try:
    from qtpy import QtCore
except ImportError:
    warn('qtpy not found; falling back to PyQt5. qtpy will be required in pyvistaqt 0.4')
    from PyQt5.QtCore import ...
else:
    from qtpy import ...
```
And then in this file do:
```suggestion
from ._qtpy.QtCore import QObject, Signal, Slot
```
And once we cut version 0.3, we can revert the `from ._qtpy import ...` lines to just be `from qtpy import ...`

But maybe this is overkill since `pyvistaqt` is still relatively new. Up to the main `pyvista` devs I guess...

larsoner (commented):
Just a question/suggestion about deprecation, otherwise LGTM

docs/index.rst
nicobako: The leading > is because it's a quote from the qtpy docs... But I'm open to changing it to something that looks nicer.

pyvistaqt/counter.py
nicobako: If the core developers think there needs to be a deprecation period, then this would be a good solution. I'm open to this.

akaszynski: I think just going ahead without the depreciation works.  Users can always downgrade to an older version, and adding in a depreciation is a bit of a pain.  Plus, installing `qtpy` should be in the `setup.py`, so it will be included upon install/upgrade.

We can always give a more descriptive error on import if we wish, but it's a bit of extra work making it backwards compatible and adding a depreciation warning as well.

larsoner:
Okay, good enough for me then!

nicobako:
Thank you @GuillaumeFavelier and @larsoner for your help! I'm really glad to see this merged into pyvistaqt!

Issue #62, opened 2020-10-14
Enable/Disable does not work?


GuillaumeFavelier:
Calling `disable()` should stop plotter interactions but the following stay interactive:

```py
import pyvista as pv
from pyvistaqt import BackgroundPlotter
p = BackgroundPlotter()
p.add_mesh(pv.Cube())
p.disable()  # <<< stop interactions here
p.show()
```

This follows https://github.com/pyvista/pyvista-support/issues/260#issuecomment-708479972

banesullivan:
Ah, we may need to override those methods on the BackgrounPlotter to make sure the qtinteractor is disabled/enabled

GuillaumeFavelier:
> make sure the qtinteractor is disabled/enabled

Do you know how to achieve that? I tried #63 without success so far

PR #63, opened 2020-10-15
Override enable/disable

GuillaumeFavelier:
This is an attempt to fix https://github.com/pyvista/pyvistaqt/issues/62

So far I override `enable()` and `disable()` in `BackgroundPlotter` with a naive:

```py
    def enable(self):
        self.renderer.enable()

    def disable(self):
        self.renderer.enable()
```

But this does not work :cry:

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -301,13 +301,13 @@ class QtInteractor(QVTKRenderWindowInteractor, BasePlotter):
     @wraps(BasePlotter.enable)
     def enable(self) -> None:
         """Wrap ``BasePlotter.enable``."""
-        self.interactor.setEnabled(True)
+        self.setEnabled(True)
         return BasePlotter.enable(self)
 
     @wraps(BasePlotter.disable)
     def disable(self) -> None:
         """Wrap ``BasePlotter.disable``."""
-        self.interactor.setDisabled(True)
+        self.setDisabled(True)
         return BasePlotter.disable(self)
 
     # pylint: disable=invalid-name,no-self-use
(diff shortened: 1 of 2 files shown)


banesullivan (changes requested):
These should go on the `QtInteractor` class:

```py
def enable(self):
    BasePlotter.enable(self)
    self.interactor.setEnabled(True)

def disable(self):
    BasePlotter.disable(self)
    self.interactor.setDisabled(True)
```

GuillaumeFavelier:
> These should go on the QtInteractor class

It works! :rocket:  Thank you so much @banesullivan !

tests/test_plotting.py
```diff
@@ -228,6 +230,12 @@ def test_qt_interactor(qtbot):
     assert render_timer.isActive()
     assert not vtk_widget._closed
 
+    # test enable/disable
```
GuillaumeFavelier, line 233:
```suggestion
    # test enable/disable interactivity
```

pyvistaqt/plotting.py
```diff
@@ -298,6 +298,18 @@ def render(self) -> None:
         """Override the ``render`` method to handle threading issues."""
         return self.render_signal.emit()
 
+    @wraps(BasePlotter.enable)
+    def enable(self) -> None:
+        """Wrap ``BasePlotter.enable``."""
+        self.interactor.setEnabled(True)
```
larsoner, line 304: don't use `self.interactor` anymore, it's just `self`

banesullivan: Ah good catch! It had been a while since I've been around this source code

banesullivan (approved):
LGTM (though I haven't been too in the loop with pyvistaqt)

larsoner:
Thanks @GuillaumeFavelier @banesullivan !

Issue #80, opened 2021-02-09
PySide testing


GuillaumeFavelier:
Now that `PySide6` has been released, it might be good time to add some Qt binding tests.

Related to https://github.com/pyvista/pyvistaqt/pull/61#issuecomment-706039944

PR #78, opened 2021-01-22
ENH: new MultiPlotter class

GuillaumeFavelier:
Mainly proof of concept `Plotter` using separated instances of `QtInteractor` to solve isolation issues with SSAO and EDL too.

image

Related to https://github.com/mne-tools/mne-python/pull/8771#issuecomment-765563603

<details>
<summary>Code snippet</summary>

```py
import numpy as np
import vtk
import pyvistaqt as pvqt
from pyvista import examples


def _enable_ssao(plotter, dataset=None, idx=None):
    lightsP = vtk.vtkLightsPass()
    opaqueP = vtk.vtkOpaquePass()
    translucentP = vtk.vtkTranslucentPass()
    volumeP = vtk.vtkVolumetricPass()

    collection = vtk.vtkRenderPassCollection()
    collection.AddItem(lightsP)

    # opaque passes
    ssaoCamP = vtk.vtkCameraPass()
    ssaoCamP.SetDelegatePass(opaqueP)

    if dataset is None:
        ssaoP = vtk.vtkSSAOPass()
        ssaoP.SetRadius(30)
        ssaoP.SetBias(0.1)
        ssaoP.SetKernelSize(32)
        ssaoP.SetBlur(True)
    else:
        bounds = np.asarray(dataset.GetBounds())
        scene_size = np.linalg.norm(bounds[:3] - bounds[3:])
        ssaoP = vtk.vtkSSAOPass()
        ssaoP.SetRadius(0.1 * scene_size)  # comparison radius
        ssaoP.SetBias(0.001 * scene_size)  # comparison bias
        ssaoP.SetKernelSize(128)  # number of samples used
        ssaoP.BlurOff()  # do not blur occlusion
    ssaoP.SetDelegatePass(ssaoCamP)
    collection.AddItem(ssaoP)

    # translucent and volumic passes
    ddpP = vtk.vtkDualDepthPeelingPass()
    ddpP.SetTranslucentPass(translucentP)
    ddpP.SetVolumetricPass(volumeP)
    collection.AddItem(ddpP)

    # finally overlays
    overP = vtk.vtkOverlayPass()
    collection.AddItem(overP)

    sequence = vtk.vtkSequencePass()
    sequence.SetPasses(collection)

    camP = vtk.vtkCameraPass()
    camP.SetDelegatePass(sequence)

    if idx is None:
        plotter.renderer.SetPass(camP)
    else:
        plotter.renderers[idx].SetPass(camP)


nefertiti = examples.download_nefertiti()
[...]

PR #81, opened 2021-02-10
MNT: Qt binding testing

GuillaumeFavelier:
Qt binding is a sensitive topic. As @nicobako explained in https://github.com/pyvista/pyvistaqt/pull/61#issuecomment-705787742:

> Switching to qtpy means that it is now up to the user to have a qtpy-compatible Qt Python binding installed

I personally see great value in `PyVistaQt` being a light Qt interface over `PyVista`. And the support of both PyQt and PySide can only be big advantage for the project. Having `qtpy` is a step in the right direction but this is still not enough. Until now, only PyQt is tested meaning that PySide is let for unofficial support. That's why I volunteer to add Qt binding testing in this PR and I started by removing `PyQt5` as the default.

Please, let me know if you have a strong opinion against this approach @pyvista/developers 

Locally, I work on a conda environment with `PySide2` since `PySide6` is not supported on [`qtpy`](https://github.com/spyder-ide/qtpy/issues/229) yet. This is a work in progress.

Closes https://github.com/pyvista/pyvistaqt/issues/80


larsoner:
My high-level comment is that I agree if PySide is meant to be officially supported it needs to be tested. I would make that happen in this PR and then in a follow-up PR work on Qt6 issues, since that's a separable issue.

GuillaumeFavelier:
Strange how those linux jobs work with `PyQt5==5.11.3` but not with `5.15.1`.

We reproduce the issues encountered in https://github.com/pyvista/pyvistaqt/pull/59 basically

GuillaumeFavelier:
I'll try updating the system from `ubuntu-16.04` to `ubuntu-18.04` but it will be my last try in this PR.

GuillaumeFavelier:
I'm actually very tempted to use Github Actions for PySide testing instead of using a matrix in Azure. I suppose it will be more convenient for the reviewers too.

GuillaumeFavelier:
Finally what I was waiting for:

```
RuntimeError: Available in PyQt5 only
```

GuillaumeFavelier:
Only this issue on Windows remains when I try to install `opengl`:

```py
Cloning into 'gl-ci-helpers'...
ParentContainsErrorRecordException: D:\a\pyvistaqt\pyvistaqt\gl-ci-helpers\appveyor\install_opengl.ps1:15
Line |
  15 |      [Net.ServicePointManager]::SecurityProtocol = 'Ssl3, Tls, Tls11,  …
     |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception setting "SecurityProtocol": "The requested security protocol is not supported."

Error: Process completed with exit code 1.
```

Any idea @akaszynski ?

larsoner:
Probably old protocols that are no longer needed (?), e.g.:

https://mcpmag.com/articles/2020/04/02/microsoft-tls-1-0-and-1-1.aspx

I would try dropping TLS 1.1 in a PR to the gl-ci-helpers repo

larsoner:
(and maybe you'll want to add a tiny Azure test for windows-latest in that repo to replicate first)

GuillaumeFavelier:
It seems that SSL3 is not supported

GuillaumeFavelier:
It is consistent with:
image

https://docs.microsoft.com/en-us/dotnet/api/system.net.securityprotocoltype?view=net-5.0#fields

GuillaumeFavelier:
Next step is adding conda.

GuillaumeFavelier:
Something in `environment.yml` introduces a dependency to `pyqt`.

GuillaumeFavelier:
It's way better now. Indeed `matplotlib` was  the culprit.

GuillaumeFavelier:

GuillaumeFavelier:
Optimally, I would prefer to merge this PR before https://github.com/pyvista/pyvistaqt/pull/78 so that I can update testing there.

larsoner (approved):
Looks like a pretty straightforward testing fix, thanks @GuillaumeFavelier

PR #76, opened 2021-01-19
MNT: Check CIs

GuillaumeFavelier:
Let's see how to make continuous integration smoother.

- [x] Fix `test_offscreen`
- [x] Fix `test_editor` on Linux
- [x] Fix `test_editor` on MacOS
- [x] Skip `test_ipython` on VTK8.1
- [x] Investigate `test_ipython` locally on `python==3.7` and `vtk==8.1.2`

Ideas:

* Add a cron-based github action runner on Linux Python3.8

PR #83, opened 2021-02-13
MNT: Add GHA schedule

GuillaumeFavelier:
This PR adds a schedule for the GitHub Actions CIs. This will help to relieve the pain of https://github.com/pyvista/pyvistaqt/pull/76 for example. By failing early, it draws attention on maintenance at least. What do you think @pyvista/developers ?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,8 @@ on:
   pull_request:
     branches:
       - '*'
+  schedule:
+    - cron: "0 4 * * *"
 
 jobs:
   pip:


akaszynski:
That's a great idea.  I recently ran into an issue when building one of my other builds that ran into an issue due to a `numpy` version change.  It's a good idea to implement daily builds and I think this would be useful for `pyvista` as well.  In fact, we need to move over some of our builds to github CI so we can reduce the time it takes to run CI by splitting up the tests between GitHub Actions and Azure Pipelines.

akaszynski (approved):

GuillaumeFavelier:
> In fact, we need to move over some of our builds to github CI

I'll be glad to help on this :+1:

akaszynski:
That would great!

I'm thinking of moving the doc build over since we can upload the artifacts of each doc-build.  Plus, there are github actions to upload docs to gh-pages (like [github-pages-deploy-action](https://github.com/JamesIves/github-pages-deploy-action)).  See https://github.com/pyansys/pymapdl/blob/master/.github/workflows/ci-build.yml

Issue #94, opened 2021-04-09
FEAT: PRE job


GuillaumeFavelier:
I think that https://github.com/pyvista/pyvistaqt/pull/83 is a step in the right direction but it might not be enough. I think a PRE job that test against PyVista **master** instead of the latest release version would be good to have. Especially, to have a bit of leeway to digest any API changes etc.

PR #95, opened 2021-04-12
MNT: Add PIP pre job

GuillaumeFavelier:
This PR adds a job that tests against PyVista master instead of the latest released version.

Closes https://github.com/pyvista/pyvistaqt/issues/94
Related to https://github.com/pyvista/pyvista/pull/1257

### DIFF
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -11,7 +11,6 @@ from qtpy.QtCore import Qt
 from qtpy.QtWidgets import QTreeWidget, QStackedWidget, QCheckBox
 from pyvista import rcParams
 from pyvista.plotting import Renderer, system_supports_plotting
-from pyvista.plotting.render_window_interactor import RenderWindowInteractor
 
 import pyvistaqt
 from pyvistaqt import MultiPlotter, BackgroundPlotter, MainWindow, QtInteractor
@@ -216,7 +215,7 @@ def test_qt_interactor(qtbot):
     vtk_widget = window.vtk_widget  # QtInteractor
 
     # check that QtInteractor.__init__() is called
-    assert_hasattr(vtk_widget, "iren", RenderWindowInteractor)
+    hasattr(vtk_widget, "iren")
     assert_hasattr(vtk_widget, "render_timer", QTimer)
     # check that BasePlotter.__init__() is called
     assert_hasattr(vtk_widget, "_closed", bool)
@@ -638,7 +637,7 @@ def test_background_plotting_close(qtbot, close_event, empty_scene):
     assert_hasattr(plotter, "app_window", MainWindow)
     assert_hasattr(plotter, "main_menu", QMenuBar)
     # check that QtInteractor.__init__() is called
-    assert_hasattr(plotter, "iren", RenderWindowInteractor)
+    hasattr(plotter, "iren")
     assert_hasattr(plotter, "render_timer", QTimer)
     # check that BasePlotter.__init__() is called
     assert_hasattr(plotter, "_closed", bool)
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -11,6 +11,7 @@ from qtpy.QtCore import Qt
 from qtpy.QtWidgets import QTreeWidget, QStackedWidget, QCheckBox
 from pyvista import rcParams
 from pyvista.plotting import Renderer, system_supports_plotting
+from pyvista.plotting.render_window_interactor import RenderWindowInteractor
 
 import pyvistaqt
 from pyvistaqt import MultiPlotter, BackgroundPlotter, MainWindow, QtInteractor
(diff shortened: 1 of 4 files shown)


GuillaumeFavelier:
I expect `pip-pre` to fail with:

```
AttributeError: BackgroundPlotter has no attribute named _add_observer
```

larsoner:
+1 for adding this, seeing a (proper) failure, and fixing the related bug!

GuillaumeFavelier:
Reproduced in https://github.com/pyvista/pyvistaqt/pull/95/checks?check_run_id=2323223467#step:9:115:

```
/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/vtkmodules/qt/QVTKRenderWindowInteractor.py in __getattr__(self, attr)
    333             return getattr(self._Iren, attr)
    334         else:
--> 335             raise AttributeError(self.__class__.__name__ +
    336                   " has no attribute named " + attr)
    337 

AttributeError: BackgroundPlotter has no attribute named _add_observer
```

GuillaumeFavelier:
Despite my few attempts on reusing the `RenderWindowInteractor` class with the current architecture, I can't make it work and it makes sense since the plotters are derived from a custom `QVTKRenderWindowInteractor` in the first place.

1) I tried to DRY as much as possible but this issue might mean a fork of how the interactions are handled in PyVista and PyVistaQt. This path means overriding all keypress events and interactions.
2) I tried making `QtInteractor` inherits from it but the base classes are not compatible.
3) Monkey-patching `self.iren` is the same as 1)

Among the less painful solutions I can think of, one would be to change the design of `RenderWindowInteractor` upstream from inheritance to composition. That way, something like:

```
self.iren = RenderWindowInteractor(self.ren_win.GetInteractor())
```

might be a compromise between successfully isolating interaction code and effectively reusing the existing `QVTKRenderWindowInteractor` without overriding everything. One could then imagine that the class can instantiate a potential `self.interactor` if `None` is provided.

GuillaumeFavelier:
Any thoughts @pyvista/developers ?

GuillaumeFavelier:
I expect each job to be red except `pip-pre` now :sweat_smile: . It seems to work on linux for the `patch/rwi` branch of PyVista. I'll be waiting for more feedback now.

GuillaumeFavelier:
Next step is to switch the `pip-pre` job to `pyvista/master` and make the CIs happy.

GuillaumeFavelier:
Hm... I'm not a fan of the latest Codecov report. Apparently we suddenly miss 3 lines somewhere:

```diff
@@            Coverage Diff             @@
##           master      #95      +/-   ##
==========================================
- Coverage   98.36%   97.95%   -0.42%     
==========================================
  Files           7        8       +1     
  Lines         612      635      +23     
==========================================
+ Hits          602      622      +20     
- Misses         10       13       +3     
```

larsoner:
In my experience Codecov is not perfect at combining reports, I would not rely on it without checking what it says (and when I do that, usually it's wrong somehow)

GuillaumeFavelier:
> Codecov is not perfect at combining reports

Okay, I was wondering because the lines are related to the `pip-pre` job.

The overall remains 97.95% so it's good enough for me.

Feel free to take a look and drop a review @larsoner if you have the time :)

adeak:
Turns out codecov shows the problems on github among the files https://github.com/pyvista/pyvistaqt/pull/95/files

I can confirm from a local test run:
```py
>     def _setup_interactor(self, off_screen: bool) -> None:
>         if off_screen:
>             self.iren: Any = None
>         else:
>             try:
                  # pylint: disable=import-outside-toplevel
>                 from pyvista.plotting.render_window_interactor import (
>                     RenderWindowInteractor,
>                 )

!                 self.iren = RenderWindowInteractor(
!                     self, interactor=self.ren_win.GetInteractor()
!                 )
!                 self.iren.interactor.RemoveObservers(
!                     "MouseMoveEvent"
!                 )  # slows window update?
!                 self.iren.initialize()
>             except ImportError:
>                 self.iren = self.ren_win.GetInteractor()
>                 self.iren.RemoveObservers("MouseMoveEvent")  # slows window update?
>                 self.iren.Initialize()
>             self.enable_trackball_style()
```
Since the `try` raises, the remaining part of the suite is not executed.

GuillaumeFavelier:
> Since the try raises, the remaining part of the suite is not executed.

This is expected for all the jobs using pyvista `0.29.0` but I added the `pip-pre` job to cover the other branch :thinking:

adeak:
Ah, right, I was using the latest pyvista release. Let me retry with pv master.

GuillaumeFavelier:
I think I found the issue, the coverage report is sent from Azure but not GitHub Actions... :sweat_smile:

GuillaumeFavelier:
It's way better now:

```diff
@@            Coverage Diff             @@
##           master      #95      +/-   ##
==========================================
+ Coverage   98.36%   98.42%   +0.05%     
==========================================
  Files           7        8       +1     
  Lines         612      633      +21     
==========================================
+ Hits          602      623      +21     
  Misses         10       10              
```


Reference: https://github.com/pyvista/pyvistaqt/pull/95#issuecomment-818678812

GuillaumeFavelier:

tests/test_plotting.py
```diff
@@ -215,7 +215,7 @@ def test_qt_interactor(qtbot):
     vtk_widget = window.vtk_widget  # QtInteractor
 
     # check that QtInteractor.__init__() is called
-    assert_hasattr(vtk_widget, "iren", vtk.vtkRenderWindowInteractor)
+    hasattr(vtk_widget, "iren")
```
larsoner, line 218: I think this has become a noop, did you mean `assert hasattr`?

GuillaumeFavelier: Yes exactly!

larsoner (approved):

larsoner:
Thanks @GuillaumeFavelier

PR #97, opened 2021-04-16
FIX: version

GuillaumeFavelier:
This PR follows https://github.com/pyvista/pyvista/pull/965 and similarly updates the version.

Related to https://github.com/mne-tools/mne-python/pull/9291#issuecomment-821186527

### DIFF
--- a/pyvistaqt/_version.py
+++ b/pyvistaqt/_version.py
@@ -1,6 +1,6 @@
 """Version info for pyvistaqt."""
 # major, minor, patch
-VERSION_INFO = 0, 3, "dev0"
+VERSION_INFO = 0, 4, "dev0"
 
 # Nice string for the version
 __version__ = ".".join(map(str, VERSION_INFO))
--- a/pyvistaqt/_version.py
+++ b/pyvistaqt/_version.py
@@ -1,6 +1,6 @@
 """Version info for pyvistaqt."""
 # major, minor, patch
-VERSION_INFO = 0, 3, 0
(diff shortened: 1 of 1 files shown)


GuillaumeFavelier:
Ready to go on my end @pyvista/developers

larsoner (approved):
Looks like the right change to me

larsoner:
... actually no it doesn't. 0.3.0 is on PyPi, so this should probably be 0.4.dev0

GuillaumeFavelier:
Thanks for checking this, I thought `0.3.dev0` would be enough

larsoner:
It's probably enough to work around the MNE issue, but the convention I'm used to with projects is that once you release 0.3.0, you update the version to 0.4.dev0 or 0.4.0.dev0 or so. That way things like looseversion comparisons make it clear you're no longer on a 0.3 variant, but an upcoming 0.4 variant

larsoner:
@GuillaumeFavelier you'll have to go on the Azure interface and restart the one failed job, I don't have permissions. If you don't have permissions we can close and reopen, or you can push an empty commit. It would be nice to just click the green button anyway since the linux failure is just a codecov upload problem, but looks like branch protection rules prohibit this

GuillaumeFavelier:
I use some kind of trick to restart it in the GitHub interface. I can't restart with DevOps itself. Usually when a job fails, the `Re-run` or `Re-run failed jobs` is available somewhere here:

image

GuillaumeFavelier:
Thanks for the reviews @larsoner !

PR #108, opened 2021-05-17
FIX: Import from root

larsoner:
https://github.com/pyvista/pyvista/pull/1310 broke things in pyvistaqt:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/8243/workflows/ab89194f-358e-4ed6-9cd1-3e9ca7fdebfa/jobs/28518

This should fix it by importing `rcParams` from `pyvista` root directly


larsoner:
This one seems pretty trivial (and I tested locally with latest 0.30 on PyPi and latest PyVista `master`) so I'll go ahead and merge

PR #109, opened 2021-05-18
MAINT: More updates

larsoner:
Write a little wrapper that allows us to use `global_theme.*` instead of using `rcParams`, but have it wrap to `rcParams` on older PyVista.

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -64,7 +64,7 @@ except ImportError:  # workaround for older PyVista
             rcParams[k] = v
 
         def __getattr__(self, k):  # noqa: D105
-            return rcParams[k]
+            return rcParams[k] if k != '__wrapped__' else None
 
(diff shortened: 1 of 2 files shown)


GuillaumeFavelier:
Thank you for working on this @larsoner 

Should we modify a CI job to test against different versions of pyvista?

Looking at the code, at least `0.29`, `0.30` and `0.30+` are maintained at the moment :sweat_smile:

larsoner:
> Should we modify a CI job to test against different versions of pyvista?

Yes probably, but I'd rather do it in another PR if possible. It's already been pretty tough to satisfy the CIs here...

GuillaumeFavelier:
I'll take care of it in a different PR then, no worries :+1:

akaszynski:
Thanks for adding this.  I kept an interface for `rcParams` to maintain backwards compatibility, as some other projects use `rcParams` and it's not great to break APIs between releases.

PR #110, opened 2021-05-18
MNT: Test against different pyvista versions

GuillaumeFavelier:
This PR follows https://github.com/pyvista/pyvistaqt/pull/109#issuecomment-843143140 and updates GitHub Actions matrix with `pyvista`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   pip:
-    name: ${{ matrix.os }} / pip / ${{ matrix.qt }} / ${{ matrix.pyvista }}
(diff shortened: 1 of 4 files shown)


GuillaumeFavelier:
The `pyvista-master` jobs should return red until https://github.com/pyvista/pyvistaqt/pull/109 is merged.

GuillaumeFavelier:
In `RangeGroup`, `setTickInterval()` expects an integer:

https://doc.qt.io/qt-5/qslider.html#tickInterval-prop

This went unnoticed and my guess is that the default value of `0.1` is just rounded to `0` (which is the default value anyway) :thinking:

GuillaumeFavelier:
I did not know about this `ResourceWarning`, maybe I'll take care of it in the future but this is low priority for now.

tkoyama010 (approved):
LGTM. It is good to test for diffrent pyvista versions. Thank you for MNT.

PR #116, opened 2021-06-07
MNT: Use wait_exposed

GuillaumeFavelier:
`waitForWindowShown` has been deprecated recently, it is now recommended to use `wait_exposed` instead:

```
>       warnings.warn(
            "waitForWindowShown is deprecated, as the underlying Qt method was "
            "obsoleted in Qt 5.0 and removed in Qt 6.0. Its name is imprecise and "
            "the pytest-qt wrapper does not raise TimeoutError if the window "
            "wasn't shown. Please use the qtbot.waitExposed context manager "
            "instead.",
            DeprecationWarning,
        )
E       DeprecationWarning: waitForWindowShown is deprecated, as the underlying Qt method was obsoleted in Qt 5.0 and removed in Qt 6.0. Its name is imprecise and the pytest-qt wrapper does not raise TimeoutError if the window wasn't shown. Please use the qtbot.waitExposed context manager instead.
```

Maybe the timeout needs to be adjusted for macOS.

### DIFF
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -147,8 +147,8 @@ def test_editor(qtbot, plotting):
     assert_hasattr(plotter, "editor", Editor)
     editor = plotter.editor
     assert not editor.isVisible()
-    editor.toggle()
-    qtbot.waitForWindowShown(editor)
+    with qtbot.wait_exposed(editor):
+        editor.toggle()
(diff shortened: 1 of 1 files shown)


larsoner (approved):
LGTM +1 for merge once CIs come back happy

PR #126, opened 2021-09-12
Improve import speed

akaszynski:
This PR switches vtk imports to either use `pyvista._vtk` or `vtkmodules`.  This results in a reduction in import time, but also permits the user to avoid including the entire `vtk` library when freezing the application with `pyinstaller` or another freezing library.

It's slightly messy because we currently allow `pyvista>=0.29.0`.  Should we raise this to `0.29.0` in the `setup.py`, we can avoid adding this in.

### DIFF
--- a/pyvistaqt/editor.py
+++ b/pyvistaqt/editor.py
@@ -2,8 +2,8 @@
 
 from typing import List
 
-import vtk
 from pyvista import Renderer
+from pyvista._vtk import vtkActor
 from qtpy.QtCore import Qt
 from qtpy.QtWidgets import (
     QCheckBox,
@@ -98,7 +98,7 @@ def _get_renderer_widget(renderer: Renderer) -> QWidget:
     return widget
 
 
-def _get_actor_widget(actor: vtk.vtkActor) -> QWidget:
+def _get_actor_widget(actor: vtkActor) -> QWidget:
     widget = QWidget()
     layout = QVBoxLayout()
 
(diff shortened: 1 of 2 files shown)


akaszynski:
Pinging @larsoner (I know you're busy).

pyvistaqt/editor.py
```diff
@@ -17,6 +16,10 @@
     QVBoxLayout,
     QWidget,
 )
+try:  # backwards compatibility with pyvista<=0.29.0
```
larsoner, line 19: Local testing suggests the first line works on 0.29.0 but not 0.28.1 from PyPi
```suggestion
try:  # backwards compatibility with pyvista<0.29.0
```

pyvistaqt/plotting.py
```diff
@@ -66,7 +65,14 @@
     QVBoxLayout,
     QWidget,
 )
-from vtk.qt.QVTKRenderWindowInteractor import QVTKRenderWindowInteractor
+try:
+    from vtkmodules.qt.QVTKRenderWindowInteractor import QVTKRenderWindowInteractor
+except ImportError:  # pragma: no cover
+    from vtk.qt.QVTKRenderWindowInteractor import QVTKRenderWindowInteractor
+try:  # backwards compatibility with pyvista<=0.29.0
```
larsoner, line 72:
```suggestion
try:  # backwards compatibility with pyvista<0.32.0
```

larsoner (commented):
Otherwise LGTM

larsoner (approved):

PR #143, opened 2022-01-14
Continuously update zoom level during pinch action

hoechenberger:
This is just a proof-of-concept @GuillaumeFavelier and I cooked up earlier today to make for a smoother pinch-zooming experience on macOS while keeping CPU utilization low.

We're refreshing at a higher frequency during the pinch action, and reset the update interval to the previous value as soon as the pinch action has ended.

We monitored CPU usage and indeed it returns to ~0 after the pinching is over.

cc @agramfort @larsoner, this makes for a **much** nicer experience with MNE `plot_alignment()` on the Mac.

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -269,6 +269,8 @@ class QtInteractor(QVTKRenderWindowInteractor, BasePlotter):
         else:
             self._setup_key_press()
 
+        self._auto_update_orig = auto_update
+
         # Make the render timer but only activate if using auto update
         self.render_timer = QTimer(parent=parent)
         if float(auto_update) > 0.0:  # Can be False as well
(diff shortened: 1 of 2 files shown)


akaszynski:
Note that if pinching can't be simulated in unit testing, we need to `pragma: no cover this to make codecov happy.

larsoner:
My suggestion here is -- rather than change the timer, which seems a bit fragile and also requires `auto_update=True` -- is to just call `Update` directly in the handler (or maybe there is a `.zoom(update=True)`). Assuming VTK/pyvistaqt is somehow smart enough to allow this sort of "update pending" call (i.e., delay the actual rendering until it's possible, for example via `paintEvent` and/or the 60 Hz limitation, at least when `vsync` is on) then this should work better. It might require some testing to see if this makes sense and/or works, but I think it's a better solution than messing with a timer.

larsoner (approved):
FWIW 7925c8f gives me:
```
  File "/Users/larsoner/python/pyvistaqt/pyvistaqt/plotting.py", line 337, in gesture_event
    int((self._auto_update_orig ** -1) * 1000.0)
ZeroDivisionError: 0.0 cannot be raised to a negative power
Fatal Python error: Aborted
```
so it was problematic (no gaurd against auto-update off). But in any case, 8787feb gives completely smooth, seemingly correct interaction on my macOS laptop. So +1 for merge from my end!

GuillaumeFavelier:
> ZeroDivisionError: 0.0 cannot be raised to a negative power

This probably happens when `auto_update=False`.

---

I added some tests for coverage, I'll merge when the CIs return green

GuillaumeFavelier:
Thank you @hoechenberger !

PR #149, opened 2022-03-01
WIP,ENH: Add support for PyQt6

GuillaumeFavelier:
This PR updates testing for PyQt6, it's not supported by `vtk` yet.

### DIFF
--- a/pyvistaqt/rwi.py
+++ b/pyvistaqt/rwi.py
@@ -1,0 +1,771 @@
+# coding=utf-8
+"""
+A simple VTK widget for PyQt or PySide.
+See http://www.trolltech.com for Qt documentation,
+http://www.riverbankcomputing.co.uk for PyQt, and
+http://pyside.github.io for PySide.
+
+This class is based on the vtkGenericRenderWindowInteractor and is
+therefore fairly powerful.  It should also play nicely with the
+vtk3DWidget code.
+
+Created by Prabhu Ramachandran, May 2002
+Based on David Gobbi's QVTKRenderWidget.py
+
+Changes by Gerard Vermeulen Feb. 2003
+ Win32 support.
+
+Changes by Gerard Vermeulen, May 2003
+ Bug fixes and better integration with the Qt framework.
+
+Changes by Phil Thompson, Nov. 2006
+ Ported to PyQt v4.
+ Added support for wheel events.
+
+Changes by Phil Thompson, Oct. 2007
+ Bug fixes.
+
+Changes by Phil Thompson, Mar. 2008
+ Added cursor support.
+
+Changes by Rodrigo Mologni, Sep. 2013 (Credit to Daniele Esposti)
+ Bug fix to PySide: Converts PyCObject to void pointer.
+
+Changes by Greg Schussman, Aug. 2014
+ The keyPressEvent function now passes keysym instead of None.
+
+Changes by Alex Tsui, Apr. 2015
+ Port from PyQt4 to PyQt5.
+
+Changes by Fabian Wenzel, Jan. 2016
+ Support for Python3
+
+Changes by Tobias Hänel, Sep. 2018
+ Support for PySide2
+
+Changes by Ruben de Bruin, Aug. 2019
+ Fixes to the keyPressEvent function
+
+Changes by Chen Jintao, Aug. 2021
+ Support for PySide6
+"""
+
+# Check whether a specific PyQt implementation was chosen
+try:
+    import vtkmodules.qt
+    PyQtImpl = vtkmodules.qt.PyQtImpl
+except ImportError:
+    pass
+
+# Check whether a specific QVTKRenderWindowInteractor base
+# class was chosen, can be set to "QGLWidget" in
+# PyQt implementation version lower than Pyside6,
+# or "QOpenGLWidget" in Pyside6
+QVTKRWIBase = "QWidget"
+try:
+    import vtkmodules.qt
+    QVTKRWIBase = vtkmodules.qt.QVTKRWIBase
+except ImportError:
+    pass
+
+from vtkmodules.vtkRenderingCore import vtkRenderWindow
+from vtkmodules.vtkRenderingUI import vtkGenericRenderWindowInteractor
+
+if PyQtImpl is None:
+    # Autodetect the PyQt implementation to use
+    try:
+        import PyQt6
+        PyQtImpl = "PyQt6"
+    except ImportError:
+        try:
+            import PySide6
+            PyQtImpl = "PySide6"
+        except ImportError:
+            try:
+                import PyQt5
+                PyQtImpl = "PyQt5"
+            except ImportError:
+                try:
+                    import PySide2
+                    PyQtImpl = "PySide2"
+                except ImportError:
+                    try:
+                        import PyQt4
+                        PyQtImpl = "PyQt4"
+                    except ImportError:
+                        try:
+                            import PySide
+                            PyQtImpl = "PySide"
+                        except ImportError:
+                            raise ImportError("Cannot load either PyQt or PySide")
+
+# Check the compatibility of PyQtImpl and QVTKRWIBase
+if QVTKRWIBase != "QWidget":
+    if PyQtImpl in ["PyQt6", "PySide6"] and QVTKRWIBase == "QOpenGLWidget":
+        pass  # compatible
+    elif PyQtImpl in ["PyQt5", "PySide2","PyQt4", "PySide"] and QVTKRWIBase == "QGLWidget":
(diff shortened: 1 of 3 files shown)


pyvistaqt/rwi.py
```diff
@@ -0,0 +1,771 @@
+# coding=utf-8
```
larsoner, line 1: Can you add a comment about where this was copied from, and under what license? Mayavi perhaps? IIRC Mayavi's RWI differs from that in the VTK repo... it would be nice to vendor here whichever parts of all RWIs work the best, to the extent that we can / know how. Some comment at the top of the file mentioning this, too, would help future maintenance and `diff`s with the other versions.

larsoner: ... and as a side note, maybe having our own version of this will someday allow us to fix #20, which to me is actually a high-priority bug. There just hasn't been a "good" way to fix it so far because we have been stuck using VTK's version: https://gitlab.kitware.com/vtk/vtk/-/issues/17953

Or maybe if #20 is somehow magically fixed by PySide6, we could just recommend that everyone move to that ASAP!

GuillaumeFavelier: Whoops, I use the one from VTK actually :sweat_smile: but I can change it actually. I did not think about the hidpi issue right away

larsoner: I am not 100% sure which is more correct. It's possible that the VTK one *was* outdated but no longer is ... or I'm just misremembering which was better. I just know that last time I looked, they were not the same.

Hence why I think it should be investigated/documented clearly here for the next time we have to think about it :)

pyvistaqt/rwi.py
```diff
@@ -0,0 +1,774 @@
+# modified from: https://gitlab.kitware.com/vtk/vtk/-/blob/master/Wrapping/Python/vtkmodules/qt/QVTKRenderWindowInteractor.py
```
larsoner, line 1: A comment about the relationship between this and the Mayavi version would be good if possible. But it could be a follow-up PR if someone wants... in which case please just add a TODO comment like `# TODO: Mayavi has a potentially different version that might be better or worse` so it's clear that we don't know which is better, and chose the VTK one just because it's the most obvious/upstream choice.

pyvistaqt/rwi.py
```diff
@@ -0,0 +1,774 @@
...
+Changes by Chen Jintao, Aug. 2021
+ Support for PySide6
+"""
+
+# Check whether a specific PyQt implementation was chosen
+try:
+    import vtkmodules.qt
+    PyQtImpl = vtkmodules.qt.PyQtImpl
+except ImportError:
+    pass
```
larsoner, line 61: These coverage comments are not really helpful or relevant if we are verbatim copying. We should add a [.coveragerc](https://github.com/mne-tools/mne-python/blob/main/.coveragerc)

GuillaumeFavelier: it's not verbatim copying though. I had to adapt it to support PyQt6

larsoner: Sounds like you need to make a PR to VTK for this, assuming it actually works :)

GuillaumeFavelier: Once I'm happy with my experiments :+1:

GuillaumeFavelier:
It's not straightforward to build an automatic testing environment that supports `PySide` or `PyQt4`. I'll drop those from my test for now.

The interesting part for `PyQt6` is:

```
  File "pyvistaqt/rwi.py", line 203, in <module>
    class QVTKRenderWindowInteractor(QVTKRWIBaseClass):
  File "pyvistaqt/rwi.py", line 274, in QVTKRenderWindowInteractor
    0:  Qt.ArrowCursor,          # VTK_CURSOR_DEFAULT
AttributeError: type object 'Qt' has no attribute 'ArrowCursor'
Error: Process completed with exit code 1.
```

GuillaumeFavelier:
The set of changes is bigger than I thought but this is much better.

Now I'm worried about:

```
Run python pyvistaqt/rwi.py
  File "pyvistaqt/rwi.py", line 130, in <module>
    from PySide6.QtWidgets import QWidget
ImportError: could not import module 'PySide6.QtGui'
Error: Process completed with exit code 1.
```

larsoner:
> Now I'm worried about:

This looks like an error with PySide6 on macOS. maybe try a different `macos-*` variant to see if it helps? If so then you could file a bug with the PySide people and we can live with only testing on an older platform for now.

> It's not straightforward to build an automatic testing environment that supports PySide or PyQt4. I'll drop those from my test for now.

Agreed these are old and we can stop supporting them

GuillaumeFavelier:
It could be related to the support on `python 3.6` and `3.7`. When I bumped to `python 3.8`, the issue was fixed.

I found a report of issues on MacOS for `PySide6 6.2.2` and `6.2.3`:

https://bugreports.qt.io/browse/PYSIDE-1797

larsoner:
If it's a bug in PySide6 I think it's fine to just test with 3.8+ until they fix it (if ever). Ideally we would document this somewhere.

Looks like you need to tell pylint and pydocstyle to ignore the copy-paste-edited file

GuillaumeFavelier:
I cannot seem to make `mypy` exclude the file. I thought `mypy --exclude 'rwi.py' pyvistaqt` would be enough.

I think the action `ricardochaves/python-lint@v1.4.0` is just outdated. The last update on https://github.com/ricardochaves/python-lint is 10 months old. If there is an update on `mypy` (e.g. bugfix), we do not have it.

GuillaumeFavelier:
From https://github.com/ricardochaves/python-lint-image/blob/master/Dockerfile

```

FROM python:3.8.0-alpine3.10

LABEL maintainer="ricardobchaves6@gmail.com"

RUN apk add --no-cache \
    gcc \
    musl-dev && \
    pip install pylint==2.7.4 \
    pylint_django==2.4.3 \
    pycodestyle==2.7.0 \
    flake8==3.9.1 \
    black==20.8b1 \
    mypy==0.812 \
    isort==5.8.0
```

Whereas `mypy 0.931` is out.

larsoner:
Can we modify the mypy repo config, e.g.:
```
[mypy]
exclude = (?x)(
    ^one\.py$    # files named "one.py"
    | two\.pyi$  # or files ending with "two.pyi"
    | ^three\.   # or files starting with "three."
  )
```
from

https://mypy.readthedocs.io/en/stable/config_file.html

GuillaumeFavelier:
This is what I just reverted @larsoner, it did not work either.

GuillaumeFavelier:
So the `pip-pre` job on linux tests against the different Qt bindings (`PyQt5`, `PyQt6`, `PySide2` and `PySide6`), run the default VTK `QVTKRenderWidgetConeExample` then tests the project.

The rest of the jobs tests against PyQt5 only so it's a bit unbalanced. I think it would be nice to have around 50/50, for instance GitHub Actions' jobs on `PySide2` and AzurePipelines' jobs on `PyQt5`.

larsoner:
This already seems like a great improvement, so +1 for merge from me already. We can always rebalance in another PR

GuillaumeFavelier:
So I have been testing this PR locally for a few days now and it works as expected. The only thing left bothering me is in this part of the code:

new code:

```py
            if "wflags" in kw:
                wflags = kw['wflags']
            else:
                wflags = Qt.WindowType.Widget  # <--- this line in particular
```

previously:
```py
            if "wflags" in kw:
                wflags = kw['wflags']
            else:
                wflags = Qt.WindowFlags()
```

So AFAIK, this line saves the Qt window flags into [`WindowFlags`](https://doc-snapshots.qt.io/qt6-dev/qt.html#WindowType-enum), a `typedef` containing the window type plus the window flags:

> The WindowFlags type is a typedef for [QFlags](https://doc-snapshots.qt.io/qt6-dev/qflags.html)<WindowType>. It stores an OR combination of WindowType values.

I found no way or creating this structure in `PyQt6` (now that everything is moved inside python `enum`s) so I simply replaced it by the default 'widget' window type and assumed zero default window flags.

GuillaumeFavelier:
This is ready from my end @pyvista/pyvistaqt-admin

larsoner:
On PyQt5:
```
>>> from PyQt5 import QtCore
>>> int(QtCore.Qt.WindowFlags())
0
>>> QtCore.Qt.Widget
0
```
So this seems backward compatible at least. It would make sense that the default WindowFlags was empty/zero anyway.

larsoner (approved):
LGTM +1 for merge! Anyone else want to look? If not, I'll merge tomorrow

larsoner:
Thanks @GuillaumeFavelier !

PR #152, opened 2022-03-09
ENH: Allow subclassing MainWindow

larsoner:
Over in MNE-Python we want to subclass the MainWindow that is instantiated by BackgroundPlotter to override event processing (https://github.com/mne-tools/mne-python/issues/9182#issuecomment-805737809) .This PR makes it possible to pass the class to use for instantiation.

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -46,8 +46,7 @@ import platform
 import time
 import warnings
 from functools import wraps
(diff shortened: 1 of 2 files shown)


larsoner:
@GuillaumeFavelier ready for review/merge from my end!

GuillaumeFavelier (approved):
Pretty straightforward changes and very good idea :+1:

PR #153, opened 2022-03-10
FIX: AttributeError MouseMoveEvent on Qt6

GuillaumeFavelier:
This PR fixes the issue with the position of the event on Qt6.

This also reveals one of the weaknesses of our testing since we do not simulate clicks. I could take the time to address this in here if I can.

Related to https://github.com/pyvista/pyvistaqt/issues/20#issuecomment-1064069790


GuillaumeFavelier:
This is ready to go from my end @pyvista/pyvistaqt-admin

PR #157, opened 2022-03-29
MNT: conda forge

GuillaumeFavelier:
`pyqt` **should not** be in the recipe of `conda-forge/pyvistaqt-feedstock` since `pyvistaqt` does not install *any* Qt API.

The goal of the PR is to help with this issue:

https://github.com/conda-forge/pyvistaqt-feedstock/pull/9#issuecomment-996413063

### DIFF
--- a/pyvistaqt/__init__.py
+++ b/pyvistaqt/__init__.py
@@ -4,11 +4,12 @@ from ._version import __version__
 try:
     from qtpy import QtCore  # noqa
 except ModuleNotFoundError as exc:  # pragma: no cover
+    _exc_msg = exc
 
     # pylint: disable=too-few-public-methods
     class _QtBindingError:
         def __init__(self, *args, **kwargs):
-            raise RuntimeError(f"No Qt binding was found, got: {exc}")
+            raise RuntimeError(f"No Qt binding was found, got: {_exc_msg}")
 
     # pylint: disable=too-few-public-methods
     class BackgroundPlotter(_QtBindingError):
--- a/pyvistaqt/__init__.py
+++ b/pyvistaqt/__init__.py
@@ -3,7 +3,7 @@ from ._version import __version__
 
 try:
     from qtpy import QtCore  # noqa
-except Exception as exc:  # pragma: no cover
+except Exception as exc:  # pragma: no cover # pylint: disable=broad-except
     _exc_msg = exc
 
     # pylint: disable=too-few-public-methods
--- a/pyvistaqt/__init__.py
+++ b/pyvistaqt/__init__.py
@@ -3,7 +3,7 @@ from ._version import __version__
 
 try:
     from qtpy import QtCore  # noqa
-except ModuleNotFoundError:
+except ModuleNotFoundError:  # pragma: no cover
 
     # pylint: disable=too-few-public-methods
     class _QtBindingError:
--- a/pyvistaqt/__init__.py
+++ b/pyvistaqt/__init__.py
@@ -1,6 +1,28 @@
 """PyVista package for 3D plotting and mesh analysis."""
-from pyvistaqt._version import __version__
-from pyvistaqt.plotting import BackgroundPlotter, MainWindow, MultiPlotter, QtInteractor
+from ._version import __version__
+
+
+try:
+    from qtpy import QtCore  # noqa
+except ModuleNotFoundError:
+    class BackgroundPlotter:
+        def __init__(self, *args, **kwargs):
+            raise RuntimeError('No Qt binding found')
+
+    class MainWindow:
+        def __init__(self, *args, **kwargs):
+            raise RuntimeError('No Qt binding found')
+
+    class MultiPlotter:
+        def __init__(self, *args, **kwargs):
+            raise RuntimeError('No Qt binding found')
+
+    class QtInteractor:
+        def __init__(self, *args, **kwargs):
+            raise RuntimeError('No Qt binding found')
+else:
+    from .plotting import BackgroundPlotter, MainWindow, MultiPlotter, QtInteractor
+
 
 __all__ = [
     "__version__",
--- a/pyvistaqt/__init__.py
+++ b/pyvistaqt/__init__.py
@@ -5,26 +5,26 @@ try:
     from qtpy import QtCore  # noqa
 except ModuleNotFoundError:
 
-    # pylint: disable=too-few-public-methods,missing-class-docstring
+    # pylint: disable=too-few-public-methods
     class _QtBindingError:
         def __init__(self, *args, **kwargs):
             raise RuntimeError("No Qt binding is found")
 
-    # pylint: disable=too-few-public-methods,missing-class-docstring
+    # pylint: disable=too-few-public-methods
     class BackgroundPlotter(_QtBindingError):
-        pass
+        """Handle Qt binding error for BackgroundPlotter."""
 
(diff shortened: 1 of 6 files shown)


GuillaumeFavelier:
What do you think @pyvista/pyvistaqt-admin ?

pyvistaqt/__init__.py
```diff
@@ -1,6 +1,34 @@
...
+from ._version import __version__
+
+try:
+    from qtpy import QtCore  # noqa
+except ModuleNotFoundError:  # pragma: no cover
+
+    # pylint: disable=too-few-public-methods
+    class _QtBindingError:
+        def __init__(self, *args, **kwargs):
+            raise RuntimeError("No Qt binding is found")
```
larsoner, line 11:
```suggestion
            raise RuntimeError(f"No Qt binding was found, got: {exc}")
```

pyvistaqt/__init__.py
```diff
@@ -1,6 +1,34 @@
 """PyVista package for 3D plotting and mesh analysis."""
-from pyvistaqt._version import __version__
-from pyvistaqt.plotting import BackgroundPlotter, MainWindow, MultiPlotter, QtInteractor
+from ._version import __version__
+
+try:
+    from qtpy import QtCore  # noqa
+except ModuleNotFoundError:  # pragma: no cover
```
larsoner, line 6:
```suggestion
except ModuleNotFoundError as exc:  # pragma: no cover
```

tests/conftest.py
```diff
@@ -1,14 +1,29 @@
...
     if NO_PLOTTING:
         pytest.skip(NO_PLOTTING, reason="Requires system to support plotting")
     yield
+
+
+@pytest.fixture()
+def no_qt():
+    """Require plotting."""
+    if _check_qt_installed():
+        pytest.skip(reason="Requires that Qt is not installed")
```
larsoner, line 28: instead of doing this, I would `monkeypatch.setattr` something to make it behave like nothing is installed. It might be as easy as `monkeypatch.setenv('QT_API', 'bad_name')` or so, since it does appear to cause a failure loally at least:
```
$ QT_API=bad_name python -c "from qtpy import QtCore"
  File "<string>", line 1, in <module>
  File "/home/larsoner/.local/lib/python3.10/site-packages/qtpy/__init__.py", line 106, in <module>
    assert API in (PYQT5_API + PYQT6_API + PYSIDE2_API + PYSIDE6_API)
AssertionError
```

GuillaumeFavelier: But it would be a different test, right? Compared to really checking the installation. Plus I think `qtpy` throws a specific `PythonQtError` when no Qt API is found, the one you mention is an failed assertion.

larsoner: ... my suggested change makes it so that regardless of whether or not the dev's system has Qt installed, the test runs and checks what it's supposed to check. What you have now requires that the system *not* have Qt in order to run, which will not be the case on almost any dev's system, rendering the test a huge pain to actually execute locally.

GuillaumeFavelier: I'll work on this, like you said, maybe there is a way to make `qtpy` behave like nothing is installed :+1:

larsoner: I think you just put a `monkeypatch.setenv('QT_API', 'bad_name')` inside the conditional here instead of a `pytest.skip` and everything should just work...

GuillaumeFavelier: But seeting `QT_API` is like requesting the wrong API :sweat_smile: 

I think I found a way by looking at `qtpy` code. It checks the internal boolean variables `PYQT6`, `PYQT5`, `PYSIDE2`, `PYSIDE6` and if none is true, it raises `PythonQtError`. It seems close enough to no Qt API found, WDYT?

larsoner: > But setting QT_API is like requesting the wrong API ￼

That is exactly the point, to make it (temporarily, via monkeypatch) fail on systems where it would otherwise pass. This is what I'm trying to accomplish...

> It checks the internal boolean variables PYQT6, PYQT5, PYSIDE2, PYSIDE6 and if none is true, it raises PythonQtError. It seems close enough to no Qt API found, WDYT?

Maybe but it seems more fragile to mess with `qtpy` internals rather than relying on a public API (QT_API) that we know we can use to make Qt appear to be missing.

I still don't understand the hesitation to route via QT_API?

GuillaumeFavelier: > I still don't understand the hesitation to route via QT_API?

Sorry if I feel picky, I just want to test a scenario as close as possible to the real situation. But it's not necessary really, as long as we keep the CI job that tests without any Qt backend installed, I'll be happy :+1:

larsoner: If you think you can do it easily/cleanly/as easily maintained via a `monkeypatch` directly to `qtpy` (or some other method) feel free. I really just want to make sure that we can emulate `qtpy` failing to import a Qt API / raising some error even on systems where it would normally succeed so that the `no_qt` tests can actually run for devs

GuillaumeFavelier:
Locally it did not raise:

```
    def test_no_qt_binding(no_qt):
        from pyvistaqt import BackgroundPlotter, MainWindow, MultiPlotter, QtInteractor
        with pytest.raises(RuntimeError, match="No Qt binding"):
>           BackgroundPlotter()
E           Failed: DID NOT RAISE <class 'RuntimeError'>
```

I think it's because `_check_qt_installed` actually imports `qtpy` before monkeypatching.

tests/conftest.py
```diff
@@ -1,14 +1,29 @@
...
     if NO_PLOTTING:
         pytest.skip(NO_PLOTTING, reason="Requires system to support plotting")
     yield
+
+
+@pytest.fixture()
+def no_qt(monkeypatch):
+    """Require plotting."""
+    if _check_qt_installed():
+        monkeypatch.setenv('QT_API', 'bad_name')
```
larsoner, line 28: Indeed, I think in addition to this you'll need to force reloading of the module somehow if possible, maybe via `importlib.reload`?

larsoner: (you can call it after the monkeypatch, then again after the `yield` and maybe this is enough)

GuillaumeFavelier: Oh nice! I did not know it was possible.

GuillaumeFavelier: But now it fails inside the `no_qt` fixture!

larsoner: That's fine I think, just `try: except: pass` it. Then the `from pyvistaqt import ...` will hit the same exception in the test.

larsoner:
I think I found a solution, I'll push...

GuillaumeFavelier:
Thank you for the fix, I have to admit it is indeed a very elaborate trickery there :smile:

larsoner:
Thanks @GuillaumeFavelier !

larsoner:
@GuillaumeFavelier feel free to cut a release and we can see if we can remove `qtpy` from the `pyvistaqt` requirements

GuillaumeFavelier:
You mean *`pyqt` right? :sweat_smile:

larsoner:
Sure, whatever was referenced in the feedstock :)

GuillaumeFavelier:
> feel free to cut a release

Done, `pyvistaqt 0.8.0` is now live! In sync with `pyvista 0.34.0` :+1:

PR #161, opened 2022-05-25
Fixed auto-update docstring in plotting.QtInteractor

### LINKED ISSUE #141: Inaccurate docstring for the 'auto_update' argument of QtInteractor, BackgroundPlotter, and MultiPlotter
The docstrings for these classes states:

```
    auto_update :
        Automatic update rate in seconds.  Useful for automatically
        updating the render window when actors are change without
        being automatically ``Modified``.
```

However, in [the implementation of `auto_update`](https://github.com/pyvista/pyvistaqt/blob/main/pyvistaqt/plotting.py#L272-L282), the value of `auto_update` is raised to the power of `-1`:
```py
twait = int((auto_update ** -1) * 1000.0)
```
meaning that `auto_update` actually specifies the number of updates per second (larger values = more updates), rather than the update rate in seconds (larger values = fewer updates).

I'm not sure if this is an issue with the docs or with the implementation?

charlessillett:
Hi, Thanks for making pyvistaqt! I'm new to contributing to open-source software, and I've made a small change to the docstring from Issue #141

Fixes #141

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -171,7 +171,7 @@ class QtInteractor(QVTKRenderWindowInteractor, BasePlotter):
         If True, enable polygon smothing
 
     auto_update :
-        Automatic update rate in seconds.  Useful for automatically
+        Number of updates per second.  Useful for automatically
         updating the render window when actors are change without
         being automatically ``Modified``.
     """


tkoyama010 (approved):
LGTM Thanks!

Issue #163, opened 2022-06-12
Bug: `mypy` cannot access `QMainWindow` methods from `MainWindow`


adam-grant-hendry:
`__init__.py` imports `MainWindow` from `plotting`. However, `MainWindow` is fully exported from `window`, not `plotting`. While not incorrect python, this causes mypy to not be able to see `QMainWindow` super class methods when creating `pyvistaqt.MainWindow` instances.

e.g.

```python
from pyvistaqt import MainWindow

window = MainWindow()
window.setWindowTitle('window')  # mypy complains window has no attribute "setWindowTitle"
```

Also `mypy<=0.961` does not yet fully support conditional imports (see [Issue 1297](https://github.com/python/mypy/issues/1297) and [1393](https://github.com/python/mypy/issues/1393#issuecomment-1153228303)).

The fix for the first item is to either use a package wildcard import or import from the appropriate modules using the `from ... import ... as ...` notation:

### Option 1: Explicit

```python
...
else:
    from .plotting import (
        BackgroundPlotter as BackgroundPlotter,
        MultiPlotter as MultiPlotter,
        QtInteractor as QtInteractor,
    )
    from .window import MainWindow as MainWindow
```

### Option 2: Implicit

```python
...
else:
    from .plotting import *
    from .window import *
```

The fix for the second item would simply be to raise the `RuntimeError` directly within the `except` clause and use the [@cjerdonek workaround](https://github.com/python/mypy/issues/1297#issuecomment-508593494):

```python
try:
    from qtpy import QtCore as _QtCore
except Exception as exc:
    QtCore = None  # pylint: disable=invalid-name
    raise RuntimeError(f'No Qt binding was found, got: {exc}') from exc
else:
    QtCore = _QtCore
    # Do imports from ``plotting`` and ``window`` here
```

**NOTE:** [PEP 484](https://peps.python.org/pep-0484/#stub-files) states

> - Modules and variables imported into the stub are not considered exported from the stub unless the import uses the `import ... as ...` form or the equivalent `from ... import ... as ...` form. (UPDATE: To clarify, the intention here is that only names imported using the form `X as X` will be exported, i.e. the name before and after `as` must be the same.)

> - However, as an exception to the previous bullet, all objects imported into a stub using `from ... import *` are considered exported. (This makes it easier to re-export all objects from a given module that may vary by Python version.)

and [Pyright indicates](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface)

[...]

PR #164, opened 2022-06-18
Use dependabot and dropping Python3.6 support

tkoyama010:
Using depandabot: https://github.com/pyvista/pyvista/pull/1972 .
Dropping Python3.6 and Python3 support: https://github.com/pyvista/pyvista/pull/1716

To solve the CI error in https://github.com/pyvista/pyvistaqt/pull/161 .

PR #170, opened 2022-06-28
Fixes Issue #163

adam-grant-hendry:
Issue #163 states failure to use redundant alias imports causes `mypy` to generate error messages about not being able to find parent methods.

The first commit in this PR adds a purposefully failing regression test, demonstrating the error indeed occurs. The second commit corrects the error and causes the test to pass.

This fixes Issue #163


adam-grant-hendry:
@tkoyama010 @banesullivan @akaszynski I think there are some Python 3.6 and Azure Pipeline tests failing here that are due for sunset pretty soon. Do you see any items I am missing in this PR that I should fix?

akaszynski:
I'm surprised we're still supporting Python 3.6. Can we drop this @pyvista/developers?

adam-grant-hendry:
@akaszynski I noticed some of the conda tests are failing because they don't have `mypy`, but it's in the `requirements_tests.txt` file for `pip`. Should I add that dependency to the `environment.yml`?

tkoyama010:
@akaszynski Absolutely. Please merge https://github.com/pyvista/pyvistaqt/pull/164/files.

adam-grant-hendry:
@akaszynski @tkoyama010 Are either of you against me adding `pre-commit` like in `pyvista` as part of this PR as well?

tkoyama010:
> Are either of you against me adding pre-commit like in pyvista as part of this PR as well?

+1 for it. Let's add it!

akaszynski:
> > Are either of you against me adding pre-commit like in pyvista as part of this PR as well?
> 
> @adam-grant-hendry +1 Let's add it!

I would prefer adding this in a separate PR as it's harder to tell which changes are simply a refactor and which changes are genuine feature changes.

.pylintrc
```diff
@@ -60,89 +60,97 @@ confidence=
...
+#         deprecated-operator-function,
+#         deprecated-urllib-function,
+#         xreadlines-attribute,
+#         deprecated-sys-function,
+#         exception-escape,
+#         comprehension-escape,
+#
+# This option was removed in PR 3571
+# See also: https://pylint.pycqa.org/en/latest/whatsnew/2/2.6/summary.html#summary-release-highlights
+#         bad-continuation,
```
tkoyama010, line 139: Do we need to save this as a comment-out?

adam-grant-hendry: No, we can remove this. I just wanted you to see why I deleted those, hence the comments.

adam-grant-hendry:
@akaszynski @tkoyama010 One note: python 3.7 only supports numpy up to version 1.21.6.

> I would prefer adding this in a separate PR as it's harder to tell which changes are simply a refactor and which changes are genuine feature changes.

Shoot, I saw this after I made the changes...any chance I could convince you to reconsider?

adam-grant-hendry:
@akaszynski @tkoyama010 I'm not sure why coverage has dropped. I think it has something to do with line length. I started by making all lengths <= 72 and then increased it to 100, but rather than increase coverage it decreased it. I'm not sure what's going on. Can you help?

Otherwise, all other tests are passing.

adeak:
> > I would prefer adding this in a separate PR as it's harder to tell which changes are simply a refactor and which changes are genuine feature changes.
> 
> Shoot, I saw this after I made the changes...any chance I could convince you to reconsider?

I'm just the peanut gallery here, because I don't developy pyvistaqt. But for pyvista, having blackening commits separately (i.e. having blackening PRs separately) means that I can use a `git-ignore-revs-file` that lets me see a blame not tainted by code churn caused by black. See https://github.com/pyvista/pyvista/pull/2684. Depending on whether current and any future pyvistaqt devs think this is useful, this would be a strong motivator to pull out the style changes into a separate PR.

adam-grant-hendry:
> > > I would prefer adding this in a separate PR as it's harder to tell which changes are simply a refactor and which changes are genuine feature changes.
> > 
> > 
> > Shoot, I saw this after I made the changes...any chance I could convince you to reconsider?
> 
> I'm just the peanut gallery here, because I don't developy pyvistaqt. But for pyvista, having blackening commits separately (i.e. having blackening PRs separately) means that I can use a `git-ignore-revs-file` that lets me see a blame not tainted by code churn caused by black. See [pyvista/pyvista#2684](https://github.com/pyvista/pyvista/pull/2684). Depending on whether current and any future pyvistaqt devs think this is useful, this would be a strong motivator to pull out the style changes into a separate PR.
> 
> (If you want to do this it's probably easier to turn this branch into a blackening branch, and creating a new branch for your #163 fixing changes.)

Are you recommending I separate concerns then and create a separate branch?

adam-grant-hendry:
> Well, like I said your choice here isn't likely to affect me now or in the future, so I'm hesitant to pressure you in any way :) Just an additional (significant) data point for your choice.
> 
> But yes, what I'm saying is that if you separate the trivial style change churn into a separate PR, then it can be ignored in git blames (both locally and on github, as of recently). This is all necessary because we squash on merge, so a complete PR becomes a single commit.

Your views are valued and suggestions valued. I leave the decision up to @akaszynski and @tkoyama010 . Well have to revert some merges, but I'm ok with a group decision. We're the ones who will have to maintain things.

adam-grant-hendry:
@adeak and loathe though I am to redo work, I agree with you that it makes sense to separate concerns. I believe @akaszynski shares the same sentiment.

adeak:
For what it's worth "we" often mix and match things in PRs, and while it's not as pure as I'd like I don't think there's anything inherently wrong with that as long as the result is reviewable, and if critical changes are not buried in an unrelated PR.

This here is a technicality that our commits are in one-to-one correspondence with PRs. If this weren't the case we could leave this as is, and list a dozen commits from this PR in the ignore-revs file.

PR #200, opened 2022-07-14
fix(qvtkinteractor): fix native widget drawing

### LINKED ISSUE #196: QPainter::begin: Paint device returned engine == 0, type: 1
### Discussed in https://github.com/pyvista/pyvistaqt/discussions/194

<div type='discussions-op-text'>

<sup>Originally posted by **adam-grant-hendry** July  8, 2022</sup>
When using `QtInteractor` (Windows 10 with PySide6)

```python
from pyvistaqt import MainWindow, QtInteractor
from qtpy import QtWidgets

class Window(MainWindow):
    def __init__(self):
        super().__init__()
        self.setWindowTitle('Example')

        self.layout_ = QtWidgets.QVBoxLayout()

        self.container = QtWidgets.QFrame()
        self.container.setLayout(self.layout_)
        self.setCentralWidget(self.container)

        self.plotter = QtInteractor(parent=self.container)
        self.layout_.addWidget(self.plotter)
        self.plotter.show()

        self.signal_close.connect(self.plotter.close)


app = QtWidgets.QApplication([])
window = Window()
window.show()
app.exec_()
```

The error:

```
QPainter::begin: Paint device returned engine == 0, type: 1
```

appears every time the screen is updated (even if `plotter.show()` is placed after `window.show()`.

From the [PySide6 QPaintEngine docs](https://doc.qt.io/qtforpython-6/PySide6/QtGui/QPaintEngine.html#PySide6.QtGui.PySide6.QtGui.QPaintEngine.Type), the error is saying `Qt` thinks I am on `X11`:

- `QPaintEngine.X11 == 0`
- `QPainEngine.Windows == 1`

After debugging in VSCode with the following `launch.json` so the debugger goes into packages:

```json
{
    "configurations": [
        {
            "name": "Python: Current File",
            "type": "python",
            "request": "launch",
            "program": "${file}",
            "console": "integratedTerminal",
            "justMyCode": false
        }
    ],
}
```

I found the error happens when `rwi.py::QVTKRenderWindowInteractor.paintEngine()` is called in response to a `Paint` event:

```python
def paintEngine(self):
    return None
```

I also noticed `pyvistaqt.rwi::QVTKRenderWindowInteractor.__init__()` sets widget attribute

```python
self.setAttribute(WidgetAttribute.WA_PaintOnScreen)
```

From the [PySide6 docs](https://doc.qt.io/qtforpython/PySide6/QtOpenGLWidgets/QOpenGLWidget.html#more), this flag is only supported on `X11` and requires returning NULL (`None`) from `paintEngine`:

image

Hence, an alternative is needed for non-`X11` systems (`Windows`, `Mac`) running `PySide` for OpenGL to be possible.

adam-grant-hendry:
- Check if system ix `X11` before setting `WA_PaintOnScreen`
- Enable native widgets by using `createWindowContainer` in `QtInteractor`

Fixes #196

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -270,6 +270,9 @@ class QtInteractor(QVTKRenderWindowInteractor, BasePlotter):
         self._first_time = False  # Crucial!
         LOG.debug("QtInteractor init stop")
 
+        window_ = QtGui.QWindow.fromWinId(self.winId())
+        self.createWindowContainer(window_, self)
+
     def _setup_interactor(self, off_screen: bool) -> None:
         if off_screen:
             self.iren: Any = None
--- a/pyvistaqt/rwi.py
+++ b/pyvistaqt/rwi.py
@@ -55,6 +55,7 @@ Changes by Chen Jintao, Aug. 2021
 Changes by Eric Larson and Guillaume Favelier, Apr. 2022
  Support for PyQt6
 """
+import os
 
 # Check whether a specific PyQt implementation was chosen
 try:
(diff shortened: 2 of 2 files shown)


adam-grant-hendry:
@larsoner Yes, let me confirm where it’s happening.

adam-grant-hendry:
@larsoner I think the segfaults were from trying to add the window container within `QVTKRenderWindowInteractor`:

```python
window_ = QtGui.QWindow.fromWinId(self.plotter.winId())
self.container = self.createWindowContainer(window_, self)
```

We may just have to teach users to add this to their `MainWindow` classes. I cannot find a good place to put this, unless we explicitly make `QMainWindow` subclass from one of `QtInteractor` or `BackgroundPlotter`.

adam-grant-hendry:
@larsoner @tkoyama010 Can I request one of you please review?

larsoner:
I should be able to review in about a week. One comment without looking at the changes though:

> We may just have to teach users to add this to their MainWindow classes. 

We use a MainWindow subclass that gets used for BackgroundPlotter:
```
pyvistaqt/window.py:10:class MainWindow(QMainWindow):
```
Maybe it could be added there? That would benefit PyVistaQt users who use BackgroundPlotter already (rather than the QtInteractor directly). This would also then test that your change works for BackgroundPlotter, and hence should work for users as well.

And +1 for modifying `docs/usage.rst:    class MyMainWindow(MainWindow):` with this advice

larsoner:
> I should be able to review in about a week. One comment without looking at the changes though:

I took a quick look and this does look good so far. Hopefully the `MainWindow` changes are not too painful to implement and just work immediately!

adam-grant-hendry:
@larsoner @tkoyama010 I was wrong: the errors occur when `QVTKRenderWindowInteractor` subclasses from anything other than `QWidget`.

On Windows, X11, and MacOS, with `Qt>5.4` the [primary paint engine for `QWidget`-based classes is the raster paint engine](https://doc.qt.io/qt-6/qpaintengine.html#details). When the raster engine is used, Windows and macOS ignore `WA_PaintOnScreen`, but since we call `winId` (which works on all platforms), [we get a native widget with the raster engine](https://doc.qt.io/qt-5/qwidget.html#native-widgets-vs-alien-widgets), so returning `None` from `QWidget::paintEngine()` doesn't cause an error.

Using an [opengl paint engine](https://doc.qt.io/qtforpython-6/PySide6/QtGui/QPaintEngine.html#PySide6.QtGui.PySide6.QtGui.QPaintEngine.Type), however, causes a problem. Since `Qt4`, [all QWidget's are double-buffered](https://doc.qt.io/archives/qt-4.7/qt4-arthur.html#widget-double-buffering), and `WA_PaintOnScreen` was added for backwards-compatibility to [allow setting a widget to use single buffering](http://web.archive.org/web/20150508183556/http://doc.qt.digia.com/4.0/qt4-intro.html#paint-events). However, single buffering will not work with `QOpenGLWidgets` as the [Qt docs clearly state](https://doc.qt.io/qt-5/qopenglwidget.html#relation-to-qglwidget):

> `QOpenGLWidget` always renders offscreen, using framebuffer objects. `QGLWidget` on the other hand uses a native window and surface.

Hence `QOpenGLWidget` is always going to be double-buffered.

The [QVTKOpenGL](https://github.com/Kitware/VTK/tree/master/GUISupport/Qt) widgets were [introduced to work with Qt's `QOpenGLWidget`](https://discourse.vtk.org/t/how-to-decide-between-qvtkopengl-native-widget/1070), but these have not been ported to Python VTK (i.e. `vtkmodules`).

In `PyQt5>5.4` and all of `PyQt6` (and their `PySide` equivalents), the `QGL` classes are gone and replaced by the `QOpenGL` classes. So, if we want to support the latest versions of `PyQt` and `PySide`, [let users fix rendering glitches on their hardware (esp. non-desktop versions)](https://github.com/Kitware/VTK/blob/ef927b8dd7fec7d740a77d0aae940028d81dcfda/Wrapping/Python/vtkmodules/qt/__init__.py#L41) we need DLL's for the [QVTKOpenGL widgets](https://github.com/Kitware/VTK/tree/master/GUISupport/Qt) ported to `vtkmodules`.

Alternatively, if those DLL's can't be ported, we could try to implement these classes using [the Python Wrappers developer notes](https://vtk.org/doc/nightly/html/md__builds_gitlab_kitware_sciviz_ci_Documentation_Doxygen_PythonWrappers.html).

larsoner:
> we need DLL's for the QVTKOpenGL widgets ported to vtkmodules.

This change would need to be made at the VTK end... and this would not be easy. Also I'm still concerned that, even if they added these, they would necessarily be specific to one flavor of Qt (at least 6 vs 5, if not even as specific as 6.x.y vs 6.x.z).

It might be worth searching the VTK issue tracker to see if anyone has brought up the idea of them shipping a Qt 6 variant of a `QVTKOpenGL` widget in `vtkmodules` -- that seems the most future compatible (moreso than Qt 5 at least), and if not, opening a new issue for discussion:

https://gitlab.kitware.com/vtk/vtk/-/issues

Feel free to ping me over there if you find an issue or open a new one (same username as GH)!

larsoner (changes requested):
Just requesting that our `MainWindow` subclass be modified if possible (and if it still makes sense)

adam-grant-hendry:
> > we need DLL's for the QVTKOpenGL widgets ported to vtkmodules.
> 
> This change would need to be made at the VTK end... and this would not be easy. Also I'm still concerned that, even if they added these, they would necessarily be specific to one flavor of Qt (at least 6 vs 5, if not even as specific as 6.x.y vs 6.x.z).
> 
> It might be worth searching the VTK issue tracker to see if anyone has brought up the idea of them shipping a Qt 6 variant of a `QVTKOpenGL` widget in `vtkmodules` -- that seems the most future compatible (moreso than Qt 5 at least), and if not, opening a new issue for discussion:
> 
> https://gitlab.kitware.com/vtk/vtk/-/issues
> 
> Feel free to ping me over there if you find an issue or open a new one (same username as GH)!

I spoke with David Gobbi (@dgobbi), who created the first `QVTKRenderWindowInteractor`, in [VTK Issue #18606](https://gitlab.kitware.com/vtk/vtk/-/issues/18606). We actually agreed making DLL's for the `QVTK` classes won't help: we would need to also wrap `PyQtX` libraries with `sip` and `PySideX` libraries with `shiboken`. Due to licensing and binding differences, this would require creating 4 separate packages (`vtk-qt5`, `vtk-qt6`, `vtk-pyside2`, and `vtk-pyside6`), which is not maintainable nor desired.

Instead, we agreed a pure python approach would be best. To that end, I've already begun writing all the `QVTK` C++ classes in pure python (using the binding guides from [the Python Wrapper notes](https://vtk.org/doc/nightly/html/md__builds_gitlab_kitware_sciviz_ci_Documentation_Doxygen_PythonWrappers.html)). I'm also adding a `Testing` subfolder with a `CMakeLists.txt` that will add `unittest`-style tests (this pattern is consistent with other such tests in the `VTK` library).

`QOpenGLWidget` is needed because the current `QVTKRenderWindowInteractor` implementation using `WA_PaintOnScreen` forces single-buffering, which causes issues. For one, when loading in a large file and displaying a progress bar dialog, both the progress bar and interactor try to draw to the screen simultaneously, causing a deadlock (i.e. blue-circle/rainbow-wheel of death). In all GUI frameworks (`OpenGL`, `Metal`, `Direct3D`, etc.), only one thread may draw to the screen at a time (i.e. the application main thread).

This is just one example, but there are several other rendering issues that support for `QOpenGLWidget` will fix. With a back buffer, we can render to it with multiple `QThreads` (see [the official Qt example](https://doc.qt.io/qtforpython-6.2/examples/example_opengl__threadedqopenglwidget.html)). Not only should this fix deadlock issues, but help us render faster and better support offscreen rendering activities.

adam-grant-hendry:
> Just requesting that our `MainWindow` subclass be modified if possible (and if it still makes sense)

Yes, it still makes sense. I should push the `QVTK` python classes in an MR to [VTK Issue #18606](https://gitlab.kitware.com/vtk/vtk/-/issues/18606) today or tomorrow. The intent of these is to replace the original `QVTKRenderWindowInteractor` (perhaps a deprecation warning will be added?), so some `pyvista` and `mayavi` code will need to change (but likely not all that much).

larsoner:
Sounds good, thanks for tackling this @adam-grant-hendry !

adam-grant-hendry:
> Sounds good, thanks for tackling this @adam-grant-hendry !

Of course, my pleasure.

The VTK team has requested I get reviewers from the `pyvista` and `mayavi` maintainers. Would you be willing to be a reviewer if I pinged you when I am ready?

larsoner:
Yes, gladly!

PR #127, opened 2021-10-05
FIX: Safer close

larsoner:
Qt callbacks should never raise an error or you get a segfault:

https://github.com/mne-tools/mne-python/pull/9803/checks?check_run_id=3796599057

```
E   subprocess.CalledProcessError: Command '['/opt/hostedtoolcache/Python/3.9.7/x64/bin/python', '-uc', "import mne; mne.viz.create_3d_figure((800, 600)); backend = mne.viz.get_3d_backend(); assert backend == 'pyvistaqt', backend"]' died with <Signals.SIGABRT: 6>.
...
Traceback (most recent call last):

  File "/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/pyvistaqt/plotting.py", line 602, in close

    self.app_window.close()

RuntimeError: wrapped C/C++ object of type MainWindow has been deleted
```
This PR just more safely wraps the handler in a try/except to avoid this.

PR #258, opened 2023-01-03
BUG: Safe app window close

larsoner:
Follow-up to https://github.com/pyvista/pyvistaqt/pull/127 which should help with this:
```
E   pytest.PytestUnraisableExceptionWarning: Exception ignored in: <function BackgroundPlotter.__del__ at 0x7f869d7a9ab0>
E   
E   Traceback (most recent call last):
E     File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pyvistaqt/plotting.py", line 741, in __del__
E       self.app_window.close()
E   RuntimeError: wrapped C/C++ object of type _MNEMainWindow has been deleted
```
Makes the code more DRY at the same time rather than duplicating the logic, as `self.close()` is already just:
```
        if not self._closed:
            # Can get:
            #
            #     RuntimeError: wrapped C/C++ object of type MainWindow has
            #     been deleted
            #
            # So let's be safe and try/except this in case of a problem.
            try:
                self.app_window.close()
            except Exception:  # pragma: no cover # pylint: disable=broad-except
                pass
```


larsoner:
This is a pretty straightforward fix and includes some CI fixes that should hopefully allow us to merge some of the dependabot PRs, so I'll mark this for merge-when-green

PR #307, opened 2023-02-06
MAINT: Set version for patch release

larsoner:
It has been almost a year since our last release. This is just a bugfix release so I think 0.9.1 is appropriate, but we could do 0.10.0 instead if people want.


banesullivan:
We don't have a release branch here, so I'm not sure why main is set to `0.10.dev0`..

banesullivan (approved):

tkoyama010 (approved):

larsoner:
Done https://pypi.org/project/pyvistaqt/0.9.1/

PR #317, opened 2023-02-16
BUG: Expose bug with scraping

larsoner:
This PR should show a failure with `pyvista` 0.38.2 / `main`: @banesullivan my test in https://github.com/pyvista/pyvistaqt/pull/306 was not complete enough. There is a remaining bug with scraping when more than one window is open. We end up with this (if I force the error to be shown rather than the misleading `rwi.py` one):
```
$ pytest tests/test_plotting.py -k scraping
_________________________________________________________________________ test_sphinx_gallery_scraping[2] __________________________________________________________________________
tests/test_plotting.py:991: in test_sphinx_gallery_scraping
    scraper(block, block_vars, gallery_conf)
../pyvista/pyvista/utilities/sphinx_gallery.py:60: in __call__
    plotter.screenshot(fname)
../pyvista/pyvista/plotting/plotting.py:5608: in screenshot
    return self._save_image(self.image, filename, return_img)
pyvistaqt/rwi.py:428: in __getattr__
    return BasePlotter.image.fget(self)
../pyvista/pyvista/plotting/plotting.py:1713: in image
    self._check_has_ren_win()
../pyvista/pyvista/plotting/plotting.py:1704: in _check_has_ren_win
    raise AttributeError('Render window is not available.')
E   AttributeError: Render window is not available.
```
This is because the `ren_win.IsCurrent()` will necessarily fail on one of the two windows. Adding a `plotter.ren_win.MakeCurrent()` before the `plotter.ren_win.screenshot()` call in `pyvista/utilities/sphinx_gallery.py` fixes this problem but to me this should not be necessary, it should probably be part of `screenshot` that it makes its own context current. But I'm not sure all the considerations that went into https://github.com/pyvista/pyvista/pull/3889 where this bug was introduced...


banesullivan:
Hmm... I agree we may want to have a context set/unset `MakeCurrent` in PyVista's `screenshot` method. I'll try to look into this later this week

larsoner:
Marking for merge when green!

PR #355, opened 2023-04-05
MAINT: Use main of setup-headless-display-action

larsoner:
Should fix the failure in #354


larsoner:
Okay looks like we actually need to fix some stuff for PySide6 :(

PR #356, opened 2023-04-06
MAINT: Use setuptools_scm for versioning

larsoner:
In many repos I work with we've moved to `setuptools_scm` for version control. It's great because when you cut a release on GitHub with a tag like 0.10.0, that will automatically show up as the version in PyPI and sdist, but the next PR that gets merged will show up as `0.11.0.dev<bunch of info>`. It thus adds info and gets rid of the need to manually tick versions -- you can just cut a release on GitHub and be done.

The only downside I've noticed is that downstream packages need to do `git+https://github.com/pyvista/pyvistaqt` to install latest `main` rather than using a `zipball`, but this is a minor drawback and easily worked around I think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,5 @@ docs/external_examples.rst
 
 # vim
 *.swp
+
+pyvistaqt/_version.py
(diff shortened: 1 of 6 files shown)


larsoner:
I'll merge this and give it a try in 0.10.0. If it ends up being problematic we'll revert!

PR #381, opened 2023-05-08
MAINT: Fix Azure macOS job

larsoner:
1. Fix the vmImage
2. Always upload coverage on Azure jobs
3. Add 0.39 PyVista job (will add to required runs)
4. Disable coverage comments (they are more noise than signal)


larsoner:
... and got rid of the `no-ci` conditionals, I don't think they're used as `[ci skip]` is more standard and is already checked by the check-skip job.

PR #395, opened 2023-06-05
MAINT: Update imports

larsoner:
Companion PR to https://github.com/pyvista/pyvista/pull/4486

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ force-exclude = '''
   | setup.py
 )
 '''
-check = true
 
 [tool.pydocstyle]
 match = '''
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -15,7 +15,6 @@ force-exclude = '''
   | setup.py
 )
 '''
-check = true
```
larsoner, line 18: This line makes it so you can't just `black pyvistaqt/plotting.py` for example to actually fix the files. I think whatever utility/CI actually wants to run just a check should use `--check` itself rather than having this live in `pyproject.toml`

larsoner:
I'm going to self-merge this so that CIs are green. But we can adjust further if needed once https://github.com/pyvista/pyvista/pull/4486 is merged

PR #396, opened 2023-06-05
DOC: Switch to pydata-sphinx-theme

larsoner:
Should be done to match pyvista's docs. Looks okay locally:

Screenshot 2023-06-05 at 11 32 13 AM

@banesullivan do we need to do anything special to get this to upload to https://qtdocs.pyvista.org/ ? Not sure how/when that gets updated

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -90,7 +90,7 @@ release = pyvistaqt.__version__
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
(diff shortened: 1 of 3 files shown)


banesullivan:
It should all be automatic! Let me know if it doesn't work after merging and I will look into it

larsoner:
@banesullivan I am not optimistic about the deployment, the latest release is 0.10.0 and the website shows 0.6.dev0 

https://qtdocs.pyvista.org/

Screenshot 2023-06-05 at 12 21 56 PM

banesullivan:
Aha, the docs CI is still trying to run on Azure. We need to move it to GH Actions. I don't have a ton of bandwidth at the moment but can try to get to this in the next week or so

larsoner:
In this repo? I can move it if it would help. Then you can just add whatever uploading stuff is necessary on top

banesullivan:
I think you could just copy/pasta this: https://github.com/pyvista/tetgen/blob/main/.github/workflows/docbuild.yml

though definitely use better actions like `peaceiris/actions-gh-pages@v3`

PR #410, opened 2023-06-28
Update VTKjs export

banesullivan:
Update the VTK.js export functionality to catch up with https://github.com/pyvista/pyvista/pull/4569


banesullivan:
I updated the GC test fixture to match the latest improvements in PyVista and looks like it's now causing failures here

larsoner:
@banesullivan I got a comment at some point about GC -- can you open a PR that updates the GC stuff here that shows the failures, then I can work to fix them at some point?

larsoner:
(Even if it doesn't cause failures, keeping the `conftest.py` changes in sync is worthwhile!)

PR #413, opened 2023-06-30
MAINT: Release automatically

### LINKED ISSUE #411: Stage release for PyVista 0.40
Some changes have been made here to be compatible with PyVista 0.40, we should stage a release, @larsoner, do you want to handle this? I feel like you've done it recently and may be able to handle it easier/faster/safer than I

larsoner:
Closes #411

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,58 @@
+# Upload a Python Package using Twine when a release is created
+
+# adapted from mne-bids-pipeline
+
+name: Build
+on:
+  release:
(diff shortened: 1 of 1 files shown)


banesullivan:
We may have been doing this in https://github.com/pyvista/pyvistaqt/blob/578f1f6c97e5449225d0cdd349f57206cb136d92/.ci/azure-pipelines.yml#L120 ?

banesullivan:
We should just strip out all of the Azure `.ci/` -- I think everything is handled on GHA at this point

PR #431, opened 2023-07-20
MAINT: Fix for latest pyvista

larsoner:
This is a hack but let's see if it works first

@banesullivan I looked a bit but couldn't find the right way to transition from rcParams to something new, any hints? Maybe it should show up with a `git grep rcParams` in `pyvista`?


larsoner:
... oh, `pv.global_theme`. We had that since 2 years ago in MNE-Python :hand_over_mouth:

PR #460, opened 2023-08-17
BUG: Show bug with SSAA

larsoner:
Companion to https://github.com/pyvista/pyvista/issues/4788


larsoner:
This is green and just shows a problem so I'll go ahead and merge. Happy to revise if people have comments, though

PR #470, opened 2023-08-31
Fix error when callback has arguments in `add_key_event`

tkoyama010:
See https://github.com/pyvista/pyvista/pull/4820 .


tkoyama010:
Close this as we only need to avoid EMPTY arguments.

PR #561, opened 2024-04-03
`PyQt6!=6.6.3` to fix integration test dependency

tkoyama010:
`PyQt6!=6.6.3` to fix integration test dependency.


larsoner:
Sorry missed this one @tkoyama010 but the equivalent change is in `main` now so we should be good to go!

tkoyama010:
Thanks!

PR #40, opened 2020-07-27
Add pylint in CI #39

tkoyama010:
See #39

PR #586, opened 2024-06-18
BUG: Fix PySide6 compatibility

larsoner:
So this is a problem:
```
$ python -c "from IPython.external.qt_for_kernel import QtGui; print(QtGui)"
<module 'PySide6.QtPrintSupport' from '/home/larsoner/python/virtualenvs/base/lib/python3.12/site-packages/PySide6/QtPrintSupport.abi3.so'>
```
The source file for `qt_for_external` says that it's kept for backward compat, so perhaps no surprise it doesn't work. I don't see any reason we need to triage based on being in ipython or not, though, given everything else comes from `qtpy` earlier anyway.

@tkoyama010 you added this in https://github.com/pyvista/pyvistaqt/pull/40/files#diff-bcab1285ff8ba2b1a25b18d87cd02a09eca4194b83196ba0c2b7939c76393d01R65, any recollection why it was necessary?

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,5 @@
+changelog:
+  exclude:
+    authors:
+      - dependabot
+      - pre-commit-ci
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -50,7 +50,6 @@ from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Type, 
 
 import numpy as np  # type: ignore
 import pyvista
-import scooby  # type: ignore
 from pyvista import global_theme
 
 try:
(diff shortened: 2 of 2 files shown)


larsoner:
... I suggest we merge this and cut a bugfix 0.11.1 for this. I added a `release.yml` to remove all the `dependabot` PRs from the autogenerated release notes because they're unhelpful.

larsoner:
I also opened https://github.com/ipython/ipython/issues/14463 but I don't think we should need `IPython` for this in any case

tkoyama010:
> @tkoyama010 you added this in https://github.com/pyvista/pyvistaqt/pull/40/files#diff-bcab1285ff8ba2b1a25b18d87cd02a09eca4194b83196ba0c2b7939c76393d01R65, any recollection why it was necessary?

It is not particularly impressive and can be removed. I apologize for not leaving a clear comment.

tkoyama010 (approved):

PR #633, opened 2024-09-10
Create security policy following Matplotlib project

tkoyama010:
### Overview

See https://github.com/matplotlib/matplotlib/pull/18537 .

### Details

- None

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,10 @@
+# Security Policy
+
+## Supported Versions
+
+Security vulnerability reports will be accepted and acted upon for all released versions.
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,10 @@
+# Security Policy
+
+## Supported Versions
+
+Security vulnerability reports will be accepted and acted upon for all released versions.
```
larsoner, line 5: Maybe a bit too broad and something like this would be better?

https://github.com/mne-tools/mne-python/blob/main/SECURITY.md#supported-versions

larsoner (commented):
Just one idea about limiting support to some N releases back rather than in perpetuity

PR #664, opened 2024-12-03
Switch conda CIs from PyQt5 to PySide6

tkoyama010:
ref https://github.com/pyvista/pyvistaqt/pull/662#issuecomment-2512217877


tkoyama010:
LGTM

PR #666, opened 2024-12-03
Remove check-jsonschema from requirements_test.txt as it is unused

tkoyama010:
Remove `check-jsonschema` from requirements_test.txt as it is unused.


tkoyama010:
LGTM

Issue #697, opened 2025-04-11
Flaky segfaults with latest PyVista


banesullivan:
We are seeing flaky segfaults on PyVista's integration tests now that we are validating with VTK 9.4

Until this can be resolved, we should limie VTK support here to <9.4

banesullivan:
Example test failure

<details>

```
Run pytest -v --cov pyvistaqt --cov-report html
============================= test session starts ==============================
platform linux -- Python 3.9.21, pytest-8.3.5, pluggy-1.5.0 -- /opt/hostedtoolcache/Python/3.9.21/x64/bin/python
cachedir: .pytest_cache
PyQt6 6.9.0 -- Qt runtime 6.9.0 -- Qt compiled 6.9.0
rootdir: /home/runner/work/pyvistaqt/pyvistaqt
configfile: pytest.ini
plugins: memprof-0.2.0, cov-6.0.0, qt-4.4.0
collecting ... collected 48 items

tests/test_plotting.py::test_create_menu_bar PASSED                      [  2%]
tests/test_plotting.py::test_setup_application PASSED                    [  4%]
tests/test_plotting.py::test_file_dialog PASSED                          [  6%]
tests/test_plotting.py::test_check_type PASSED                           [  8%]
tests/test_plotting.py::test_mouse_interactions PASSED                   [ 10%]
tests/test_plotting.py::test_ipython PASSED                              [ 12%]
tests/test_plotting.py::test_depth_peeling PASSED                        [ [14](https://github.com/pyvista/pyvistaqt/actions/runs/14395559093/job/40370516185#step:11:15)%]
tests/test_plotting.py::test_off_screen PASSED                           [ 16%]
tests/test_plotting.py::test_smoothing PASSED                            [ 18%]
tests/test_plotting.py::test_counter PASSED                              [ 20%]
tests/test_plotting.py::test_subplot_gc[True] PASSED                     [ 22%]
tests/test_plotting.py::test_subplot_gc[False] PASSED                    [ 25%]
Fatal Python error: Segmentation fault

Thread 0x00007fb3dec006c0 (most recent call first):
  File "/opt/hostedtoolcache/Python/3.9.21/x64/lib/python3.9/site-packages/psutil/_common.py", line 766 in open_binary
  File "/opt/hostedtoolcache/Python/3.9.21/x64/lib/python3.9/site-packages/psutil/_pslinux.py", line 362 in virtual_memory
[...]

PR #698, opened 2025-04-11
Limit VTK to <9.4

banesullivan:
Per #697 

cc @user27182 @tkoyama010 

Proposing we limit VTK support here until we can figure out why there are flaky segfaults on VTK 9.4 with pyvistaqt

As long as this is on `main` we'll be good upstream in PyVista's integration tests so we don't even need to go through a release cycle for this

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setup(
     install_requires=[
         'pyvista>=0.32.0',
         'QtPy>=1.9.0',
+        'vtk<9.4',
     ],
     package_data={'pyvistaqt': [
         os.path.join('data', '*.png'),


banesullivan:
Uh oh. These segfaults are happening with VTK 9.3.1 too... Now I'm really confused

user27182:
> Uh oh. These segfaults are happening with VTK 9.3.1 when run against PyVista's `main` branch. This means the breakage is in PyVista itself

The issues with PyVistaQt integration tests were not a problem until https://github.com/pyvista/pyvista/pull/6879 was merged. I think there is still value in limiting vtk < 9.4 here. 

Perhaps what you're seeing with segfaults with vtk 9.3.1 is a less frequent segfault, e.g. we still sometimes periodically get segfaults here and there with some tests across different OS and vtk versions with the PyVista CI, independent of PyVistaQt. Maybe, just by coincidence, this is what you're seeing here.

banesullivan:
Seems the segfault is consistently happening here now

larsoner:
If it's only on PyQt5 I think we can maybe live with it, especially if it's only for a test or two. Let me try another PR where I add a skip or something to see if I can get it "working"...

larsoner:
Doesn't seem isolated to PySide2, I can replicate locally with PySide6. So I think it must have been some change in PyVista :(

user27182:
> Doesn't seem isolated to PySide2, I can replicate locally with PySide6. So I think it must have been some change in PyVista :(

Can we determine if the breakage was caused by merging https://github.com/pyvista/pyvista/pull/6879? Maybe `git bisect`?

larsoner:
Just ran one, indeed it's due to 6cafaf553d166358b3618a0a81ebe514c401cee3 / #6879

larsoner:
... you can rapidly replicate with
```
pytest tests/test_plotting.py -k editor
```
it fails in ~1s, or it succeeds. At least on PySide6 + VTK 9.3.1 here

larsoner:
... and I took a bit of a shot in the dark but this seems to have fixed it:
```
git checkout a43b10fcf846b339d5475404fb8f3cafe63fcf32  pyvista/plotting/renderer.py
```
so at least we know where in the `diff` to look for the problem (maybe!)

user27182:
> ... and I took a bit of a shot in the dark but this seems to have fixed it:
> 
> ```
> git checkout a43b10fcf846b339d5475404fb8f3cafe63fcf32  pyvista/plotting/renderer.py
> ```
> 
> so at least we know where in the `diff` to look for the problem (maybe!)

Those changes deal with the axes widget, which was initially patched with a hack to make it work with vtk 9.4.0... might be worth trying to remove the hack as it's not needed for 9.4.2, maybe that's part of the issue?

larsoner:
I think I might have isolated it, will open a PR shortly

PR #726, opened 2025-07-17
Check that `axes_widget` is truthy

user27182:
Upstream as part of https://github.com/pyvista/pyvista/pull/7716, setting dynamic attributes on the plotter will no longer be allowed. As such, `BasePlotter.axes_widget` is set to `None` during init. This means that it's no longer sufficient to check if the attribute exists, and it's necessary to also check it's not None.

I am seeing integration test failures here:
https://github.com/pyvista/pyvista/actions/runs/16335129476/job/46145621302?pr=7716#step:8:1177
I'm pretty sure they'll all be resolved by this PR

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -15,13 +15,13 @@
 <script>
   (function (i, s, o, g, r, a, m) {
     i["GoogleAnalyticsObject"] = r;
-    (i[r] =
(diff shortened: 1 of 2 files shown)


tkoyama010 (approved):
LGTM

larsoner:
Thanks @user27182 !

PR #735, opened 2025-08-13
CI: Migrate GitHub Actions workflows to use uv

tkoyama010:
## Summary
- Replace pip with uv for faster and more reliable package installation across all GitHub Actions workflows
- Update ci.yml to use uv for all three jobs (pip-pre, pip, conda)  
- Update docbuild.yml to use uv for documentation builds
- Update release.yml to use uv for package building

## Benefits
- Faster dependency resolution and installation
- Better caching and lock file support
- More reliable builds with consistent dependency versions
- Reduced CI runtime


banesullivan:
Closing this after #816 

There is now preferential support for UV on main but I think the CI is pretty clean and theres not much of a benefit of switching to UV at this time

PR #810, opened 2026-04-02
FIX: Fix for bad interaction on macOS + 6.10+

larsoner:
@nchampag I decided to go with a variant of the simple fix from https://gitlab.kitware.com/vtk/vtk/-/merge_requests/12956/diffs?commit_id=3328ae95f80821baf01b8625cebd293e1e55a90f, can you see if it works for you? Worked for me on macOS M1 with PySide6 6.11.0 just now.

I'd like to merge and cut a release with a fix ASAP since this has been a bad bug for a while now 😬 

Closes #793

### DIFF
--- a/pyvistaqt/rwi.py
+++ b/pyvistaqt/rwi.py
@@ -56,6 +56,8 @@ Changes by Eric Larson and Guillaume Favelier, Apr. 2022
  Support for PyQt6
 """
 
+import sys
+
 # Check whether a specific PyQt implementation was chosen
 try:
     import vtkmodules.qt
@@ -127,6 +129,7 @@ if PyQtImpl == "PySide6":
     from PySide6.QtCore import QObject
     from PySide6.QtCore import QSize
     from PySide6.QtCore import QEvent
+    from PySide6.QtCore import __version__ as QT_VERSION
 elif PyQtImpl == "PyQt6":
     if QVTKRWIBase == "QOpenGLWidget":
         from PyQt6.QtOpenGLWidgets import QOpenGLWidget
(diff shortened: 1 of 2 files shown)


nchampag:
@larsoner, this modification does not completely fix the problem. Initial renders are good, but glitches may happen during resizing events, switching between tabs/windows, etc.

qt_view

larsoner:
Ahh, and the `__doPaintEvent` workaround doesn't have these issues?

nchampag:
No.

larsoner:
Okay pushed what I think is your version of the fix... work okay for you @nchampag ?

larsoner:
... I'll probably have to at a minimum test on a Linux machine tomorrow before merge to make sure nothing breaks there. If it does, I guess I could add a conditional for the system being macOS (and maybe Qt 6.10/6.11?) whenever we want to set the value to `False`.

nchampag:
I only tested it on macOS, so testing it on other OSes is a good idea. I haven't, yet, which is why the "fix" is quite simple.

nchampag:
Here is a simple VTK example where a QVTKRenderWindowInteractor has been inserted into a QTabWidget. The second tab contains a blank widget. The glitch occurs when switching between the tabs.

The glitch occurs for this [fix](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/12956/diffs?commit_id=3328ae95f80821baf01b8625cebd293e1e55a90f), where the initial render happens. However, switching to the second tab and back to the first reveals a blank screen.

The glitch does not occur for the `__doPaintEvent`  fix.

[vtkInteractorTab.py](https://github.com/user-attachments/files/26449748/vtkInteractorTab.py)

larsoner:
Pushing a commit to make it only set this value to `False` on macOS when the Qt version is `>=6.10`. I hope this doesn't create problems if the Qt folks fix things at their end. But if it does, at least it'll be easy to constrain the version and cut a quick release.

Will test to make sure nothing is broken on Linux tomorrow, merge, and cut a release!

larsoner:
... and @nchampag I can verify that with your example and

1. vanilla VTK it freezes
2. changing the import to `from pyvistaqt.rwi import QVTKRenderWindowInteractor` on this PR it works

changing tabs and resizing seems to work fine as well

larsoner:
Okay, verified things are still okay on Linux (no surprise since the logic should be unchanged there effectively). I'll merge and cut a release. Thanks for your suggested fix and testing @nchampag !

PR #842, opened 2026-07-02
Improve typing, check with ty, and use prek

### LINKED ISSUE #118: Update doc after qt backend transition
PyVistaQt has been slowly transitioning from PyQt5 to a qt-backend agnostic interface but now the transition over so the documentation (`usage.rst` and `README.md`) needs to be updated. A contribution to make clear that there is no default Qt backend is welcome.

**Suggestion**

- Add an **installation** section

**Context**

Discussion about allowing different Qt bindings started slightly before `0.2` (https://github.com/pyvista/pyvistaqt/issues/8) and the selected solution was to use `qtpy` (https://github.com/pyvista/pyvistaqt/pull/12, https://github.com/pyvista/pyvistaqt/pull/61). Then `qtpy` lived with the PyQt5 during this minor version until PySide testing was introduced in https://github.com/pyvista/pyvistaqt/pull/81. Finally, after another minor version of testing, `0.4` became the first version of PyVistaQt to reach this goal.

cc @adeak

### LINKED ISSUE #162: Feature Request: Add a py.typed marker
`pyvistaqt` has some partial typings, but no `py.typed` marker or stubs, so `mypy` is not able to recognize the type hints in the subpackages that have them since `pyvistaqt` is not `PEP-561` compliant.

It would be great to add a `py.typed` marker to the package: [Creating PEP 561 compatible packages](https://mypy.readthedocs.io/en/latest/installed_packages.html#creating-pep-561-compatible-packages)

NB: 

For `PyQt5`, stubs are available as a pip package:

```bash
$ pip install pyqt5-stubs
```

For `PyQt6`, they can be downloaded from this GitHub repo:

```bash
$ pip install git+https://github.com/python-qt-tools/PyQt6-stubs.git
```

For `PySide2`, they can be downloaded from pip:

```bash
$ pip install PySide2-stubs
```

and for `PySide6`, the package ships with a script `pyside6-genpyi` that generates the stubs:

```bash
$ pyside6-genpyi all
```

As for `PyQt4`, and `PySide`, I do not know as those are pretty old.

### LINKED ISSUE #163: Bug: `mypy` cannot access `QMainWindow` methods from `MainWindow`

### LINKED ISSUE #169: FR: Clarification of Packages Used for Testing
Please be aware of this issue that I have submitted to `pytest-qt`:

[BUG: Multiple PyQt/PySide Versions Not Supported #437](https://github.com/pytest-dev/pytest-qt/issues/437)

It would be good to add to our `CONTRIBUTING.md` documentation that testing will fail if multiple `PyQt`/`PySide` packages are installed at the same time.

larsoner:
1. Round out our typing and add `py.typed`
2. Switch to `ty` for type-checking (much faster)
3. Actually enforce type-checking (no `ignore_errors=True`)
4. Use `prek` for pre-commit hook (pre-commit is slower, and more importantly `pre-commit.ci` complains about the size of the installed env because `vtk` is pulled in by `pyvista`); note that locally `pre-commit` can still be used
5. Simplify `rwi.py` using `qtpy` (deviates more from upstream, but simplifies things a lot)
6. Simplify our import structure by letting `qtpy` raise a suitable error if no binding can be imported
7. Update `pyvista` minimum to ~2 years ago
8. Remove PyQt5 (EOL) from CI and official support
9. Adjust CIs accordingly based on ☝️ 

At some point we might want to force `QWidget` as the `QVTKRWIBaseClass` to simplify code. Not sure if it would affect https://github.com/pyvista/pyvistaqt/pull/200 or not. It would simplify the typing at least. 

Will need `autofix.ci` enabled for this repo and `pre-commit.ci` disabled. I disabled `pre-commit.ci` but don't have permissions to add `autofix.ci` -- @banesullivan can you add it for me? Or if there is some PyVista bot that can push commits to a forked PR, that would work, too (it's what `autofix.ci` is used for here)

Developed with Claude Opus 4.8 and Fable 5.0. Changes checked manually by me.

Closes #169
Closes #163
Closes #162
Closes #118


autofix-troubleshooter:
Hi! I'm the <a href="https://autofix.ci">autofix logo<strong>autofix.ci</strong></a> troubleshooter bot.

It looks like you correctly set up a CI job that uses the autofix.ci GitHub Action, but the [**autofix.ci GitHub App**](https://github.com/apps/autofix-ci/) has not been installed for this repository. This means that autofix.ci unfortunately does not have the permissions to fix this pull request. If you are the repository owner, please install the app and then restart the CI workflow! 😃

larsoner:
@banesullivan never mind I was able to install the app actually!

.pre-commit-config.yaml
```diff
@@ -1,12 +1,20 @@
 repos:
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v2.1.0
+  - repo: https://github.com/astral-sh/ty-pre-commit
+    rev: v0.0.55
```
larsoner, line 3: Intentionally behind `0.56` so I can `workflow_dispatch` the `autofix.ci` after merge to `main` to ensure it tries to update properly (it did for `mne-kit-gui` where I adapted the YAML from)

larsoner:
@tkoyama010 this is ready for review/merge from my end

tkoyama010 (approved):
LGTM

larsoner:
Okay I'll cut a 0.12 tomorrow unless there are objections!

PR #843, opened 2026-07-02
Fix bug with callback message formatting

larsoner:
Coupled with https://github.com/pyvista/pyvista/pull/8803, makes error messages in event handlers propagate properly. Diagnosed and fixed with Claude Opus 4.8 (and it was really tricky!).


larsoner:
Opened https://github.com/python/cpython/issues/152907 to track upstream

PR #90, opened 2021-03-11
WIP,ENH: MultiPlotter improvements

GuillaumeFavelier:
This PR adds a `margin` parameter to the `MultiPlotter`:

margin=True | margin=False
--|--
image | image

<details>

```py
import pyvista as pv
from pyvistaqt import MultiPlotter

mp = MultiPlotter(nrows=2, ncols=2, menu_bar=False, toolbar=False, margin=False)
mp[0, 0].add_mesh(pv.Sphere())
mp[0, 1].add_mesh(pv.Cylinder())
mp[1, 0].add_mesh(pv.Cube())
mp[1, 1].add_mesh(pv.Cone())
```

</details>

---

ToDo:
- [ ] Testing/Coverage

PR #844, opened 2026-07-09
Deprecate MultiPlotter

### LINKED ISSUE #480: DOC: MultiPlotter vs BackgroundPlotter
I think the motivations behind these classes should be documented more clearly. From our doc I couldn't tell:

https://qtdocs.pyvista.org/api_reference.html#pyvistaqt.BackgroundPlotter

I had to go back to https://github.com/pyvista/pyvistaqt/pull/78 to understand (what I think is) the motivation. For example is the class really needed if https://github.com/pyvista/pyvista/issues/4939 is solved?

### LINKED ISSUE #588: app_window property returns misleading value in a MultiPlotter setting
In a `MultiPlotter`, the `BackGroundPlotter.app_window` property of the children does not contain the application window but a child (or more precise a grandchild)-window of the application window.
 
It seems the application consists of a `Mainwindow` that is parent to a `QWidget` that is parent to the multiple plotters. So each `BackgroundPlotter` in the `MultiPlotter` has its own `MainWindow` which are nested in the application. 

I would have expected that: 
* Either `BackGroundPlotter.app_window` always returns the application window independent whether it is nested in a `MultiPlotter` 
* or  `MultiPlotter` would have a `MultiPlotter.app_window` that contains the reference to the global application window. 

Maybe adding a `qt_window` in the BackGroundplotter which in the MultiplotterCase contains the child-widgets, and for the single-case also contains the app-window and setting the correct `app_window` could resolve it - however that would be a breaking change if someone already relies on the `app_window` to return the child-widgets. 

Some code to highlight the issue: 
```python
import pyvista as pv
from pyvistaqt import BackgroundPlotter, MultiPlotter
from PySide6 import QtGui

print("BackGroundPlotter")
plotter = BackgroundPlotter()
it = plotter.app_window
print(f'plotter.app_window: {str(it)}')
while it.parent():
    print(it.parent())
    it = it.parent()
print('done')

print("MultiPlotter")
mplotter = MultiPlotter(nrows=1, ncols=2)
plotter = mplotter[0, 0]
it = plotter.app_window
print(f'plotter.app_window: {str(it)}')
while it.parent():
    print(it.parent())
    it = it.parent()
print('done')
```

Which outputs: 
```
BackGroundPlotter
plotter.app_window: <pyvistaqt.window.MainWindow(0x600000325d40) at 0x10a4c7900>
done
MultiPlotter
plotter.app_window: <pyvistaqt.window.MainWindow(0x600000336b80) at 0x12f4eef00>
<PySide6.QtWidgets.QWidget(0x600000336820) at 0x12f4ec6c0>
<pyvistaqt.window.MainWindow(0x6000003370c0) at 0x12f48bf00>
done
```

larsoner:
Closes #588
Closes #480
Closes #90

### DIFF
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -934,6 +934,12 @@ class MultiPlotter:
     interactive window with multiple plotters without
     blocking the main python thread.
 
+    .. note::
+        ``MultiPlotter`` is deprecated and will be removed in 0.13. Use
+        ``BackgroundPlotter`` with its ``shape`` argument instead, or embed
+        multiple ``QtInteractor`` instances directly in your own ``QLayout``
+        for full control.
+
     Parameters
     ----------
     app : optional
@@ -972,6 +978,14 @@ class MultiPlotter:
         **kwargs: Any,  # noqa: ANN401
     ) -> None:
         """Initialize the multi plotter."""
+        warnings.warn(
+            "MultiPlotter is deprecated and will be removed in 0.13. Use "
+            "BackgroundPlotter with its `shape` argument instead, or embed "
+            "multiple QtInteractor instances directly in your own QLayout "
+            "for full control.",
+            FutureWarning,
+            stacklevel=2,
+        )
         _check_type(app, "app", [QApplication, type(None)])
         _check_type(nrows, "nrows", [int])
         _check_type(ncols, "ncols", [int])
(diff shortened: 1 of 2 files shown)


larsoner:
Trivially implements what we discussed in #588 so self-merging!

Issue #445, opened 2023-08-03
Does not work on Wayland: X Error of failed request:  BadWindow (invalid Window parameter)


YuXHe15:
Hi
I was trying the small example of pyvistaqt, namely
```
import pyvista as pv
from pyvistaqt import BackgroundPlotter
sphere = pv.Sphere()
plotter = BackgroundPlotter()
plotter.add_mesh(sphere)
```
while I got error:

> X Error of failed request:  BadWindow (invalid Window parameter)
>   Major opcode of failed request:  12 (X_ConfigureWindow)
>   Resource id in failed request:  0x3
>   Serial number of failed request:  7
>   Current serial number in output stream:  8

After checking line by line I found it happends after creating the instance plotter of BackgoundPlotter()

My system is Ubuntu 22.04.2 LTS, using python 3.10

YuXHe15:
I did a little research around stackoverflow and one answer mentioned that it could happened when the window id passed is not correct, which can be found here:
[https://stackoverflow.com/questions/34704542/in-linux-get-window-class-raise-x-error-of-failed-request-badwindow-invalid](url)
I don't know if it would be helpful

mscheltienne:
Same problem with Ubuntu 24.04 LTS which defaults to Wayland.

EmmanuelMess:
A warning should be added to the documentation.

larsoner:
@EmmanuelMess PR welcome to improve the docs!

cbrnr:
Now that almost all Linux distributions default to Wayland and have removed X11 (yes, it is still possible to use XWayland, so setting `export QT_QPA_PLATFORM=xcb` still works), I wanted to ask if there are any plans to support Wayland natively.

I asked Claude Sonnet 4.6, and the change seems doable, but not trivial:

### Wayland Support in PyVistaQt: A Technical Assessment

The short answer is: yes, it's technically possible, but it requires a significant architectural change — not a simple patch. Here's a thorough breakdown of why, and what it would take.

### Root Cause of the Failure
The error `X Error of failed request: BadWindow` is a smoking gun. It reveals the core problem: on X11, Windows, and macOS, applications use `winId()` to get access to a native windowing system surface — an integer ID meaningful to the OS — in order to hand rendering control of that surface to an external toolkit (VTK in this case).

[...]

PR #847, opened 2026-07-09
Check for wayland early

larsoner:
While we look for a real fix for #445, we can a) try to create QApplication using `xcb` if one isn't specified, and b) raise an error if people try to use QtInteractor on Wayland.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -71,10 +71,16 @@ sphere.
 
 .. important::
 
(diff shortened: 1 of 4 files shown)


akaszynski:
If 848 is merged, will this still be needed?

larsoner:
Nope! #848 has a closes line for this PR in its description

PR #848, opened 2026-07-10
Refactor RWI to use QOpenGLWidget

### LINKED ISSUE #102: Flashing issue when resizing QtInteractor.
### **Epilepsy warning**

When resizing a pyqt5 window, the QtInteractor window flashes.

https://user-images.githubusercontent.com/70919881/116622816-47dbec00-a913-11eb-8f78-e5bb370dfe92.mov

### LINKED ISSUE #128: Flickering with QSplitter
I have this bug for a long time, but only narrowed it to QtInteractor recently.

From the moment that I have QtInteractor inside a QSplitter, resizing the sections with QSplitter causes heavy flickering on the other widgets in the other sections.

For example, I have QtInteractor  on the left and a form on the right.  Normally, only the left side of the QLineEdit is resizing, and the right side is static, tied to the frame that is not moving. What happens is the right side of any widget (QLineEdit, QListWidget, etc.) is moving forth and back when resizing, causing visual artifacts. QT is drawing over widgets that would normally never be allowed to. It only happens when QtInteractor  is inside the splitter.

I suspect it may have to do with using QWidget instead of OpenGL. But it could be just on my side. Has anybody used QSlitter with success? Anything strange happening when moving fast the handle?  Especially when there are a lot of widgets?

### LINKED ISSUE #196: QPainter::begin: Paint device returned engine == 0, type: 1

### LINKED ISSUE #445: Does not work on Wayland: X Error of failed request:  BadWindow (invalid Window parameter)

### LINKED ISSUE #846: Segfault at teardown when a QtInteractor window is destroyed without being closed
Split off from the multi-window freeze fix (#762). That PR fixes the interactive freeze; this is a separate, pre-existing teardown crash.

### Symptom

With ≥2 `QtInteractor` windows that are never explicitly closed (no `signal_close → plotter.close()` wiring), the process intermittently segfaults at exit. Real GPU only — does not reproduce under offscreen/`xvfb` software GL.

<details>
<summary>Minimal reproduction (run on a real display, not offscreen)</summary>

```python
"""Run a few times; intermittently segfaults at exit on a real GPU (NVIDIA)."""
import sys

from qtpy import QtWidgets, QtCore

import pyvista as pv
from pyvistaqt import QtInteractor


def make_window(title, mesh):
    win = QtWidgets.QMainWindow()
    win.setWindowTitle(title)
    frame = QtWidgets.QFrame()
    layout = QtWidgets.QVBoxLayout(frame)
    plotter = QtInteractor(frame)  # never closed -> parent.destroyed calls close() at teardown
    layout.addWidget(plotter.interactor)
    win.setCentralWidget(frame)
    plotter.add_mesh(mesh)
    win.resize(300, 250)
    win.show()
    return win, plotter


app = QtWidgets.QApplication(sys.argv)
app.setQuitOnLastWindowClosed(False)

win1, _ = make_window("Window 1", pv.Sphere())
win2, _ = make_window("Window 2", pv.Cube())

QtCore.QTimer.singleShot(800, win1.close)
QtCore.QTimer.singleShot(1600, app.quit)

app.exec() if hasattr(app, "exec") else app.exec_()
print("reached end of script")  # reached even when it segfaults during teardown
```

</details>

### Root cause

`rwi.py` connects the parent's `destroyed` signal to `self.close` (`DirectConnection`):

```python
self.parent().destroyed.connect(self.close, ConnectionType.DirectConnection)
```

For a never-closed interactor this runs GL-resource release *inside* `~QWidget()`, when the shared X display / GL context is often already gone:

```
QWidget::~QWidget() → destroyed() → QtInteractor.close() → BasePlotter.close()
  → vtkViewport::RemoveAllViewProps() → vtkActor::ReleaseGraphicsResources()
    → vtkXOpenGLRenderWindow::MakeCurrent() → glXMakeCurrent(...) → SIGSEGV
```

Flaky (~1–3/10) because it's a teardown-order race between the two windows and the X connection.

### The obvious fixes make it worse

The invariant is: release each render window's GL resources once, against its own still-valid context, never touching a sibling's. Batched cleanup violates that:

| Approach | Crashes (10 runs, NVIDIA) |
|---|---|
| baseline (never close) | ~1–3 / 10 |
| close both plotters at `aboutToQuit` | 10 / 10 |
| `Finalize()` both render windows at `aboutToQuit` | 10 / 10 |
| documented pattern (per-window `closeEvent → plotter.close()`) | 0 / 12 |

### Notes

- Reliable workaround exists today: close each plotter while alive (`signal_close → plotter.close()`, or `BackgroundPlotter`, which does this automatically).
- A real fix likely means releasing GL resources on close/hide (context valid) and making the `destroyed`-time `close()` a bookkeeping-only no-op with no `MakeCurrent` — delicate, needs a cross-driver/platform (NVIDIA/Mesa/macOS/Windows, PyQt/PySide) test matrix.

Investigated with Claude Opus 4.8

larsoner:
As a potential solution to #196 and I asked Claude to *translate* the `QVTKOpenGLNativeWidget` code from C++ to Python. It did it... then it took a lot of iteration to make it display properly. In the end it reused a lot of the mechanics from `rwi.py` to do translations that otherwise live in a `RenderWindowAdapter` in VTK C++, which I think should work. I think this is a good forward-compatible solution -- it uses the same underlying architecture that VTK does for its native Qt widget nowadays. It works with Wayland natively. :rocket: 

Long-standing bugs fixed:

- No `WA_PaintOnScreen` (#200, #196)
- SSAA tests work on macOS now
- Things not working on Wayland (#445)
- Teardown segfault when interactors are destroyed without being closed (GL resources are now released per-widget             via the context's `aboutToBeDestroyed` signal)
- Flicker/flashing during resize and in splitters (need to check!)

Behavior changes:

- Quad-buffer stereo (CrystalEyes) support is dropped
- The GL surface format is installed process-wide via `QSurfaceFormat.setDefaultFormat()`; per-widget `setFormat()` is not supported (it breaks compositing on Wayland)
- macOS VMs without GPU acceleration: Qt disables OpenGL process-wide, real Apple GPUs are unaffected (C++ `QVTKOpenGLNativeWidget` has the same limitation)

                                          

TODO:

- [x] Try locally
- [x] Check MNE-Python (https://github.com/mne-tools/mne-python/pull/14045)
- [x] docs/index.rst:76-83 still tells Wayland users to set QT_QPA_PLATFORM=xcb
- [x] Check if there are lessons to learn from https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9443

A related (abandoned?) effort is in https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9443.

Closes #847
Closes #846
Closes #445
Closes #200
Closes #196
Closes #128
Closes #102

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -17,7 +17,7 @@ Overview
    :alt: MIT License
 
 The Python package ``pyvistaqt`` extends the
-functionality of ``pyvista`` through the usage of *Qt*. 
+functionality of ``pyvista`` through the usage of *Qt*.
 Since *Qt* applications operate in a separate thread than VTK,
 you can simultaneously have an active VTK plot and a non-blocking Python session.
 
@@ -68,23 +68,6 @@ sphere.
 
     plotter = BackgroundPlotter()
     plotter.add_mesh(sphere)
-
-.. important::
-
-   On Linux/Wayland, Qt may currently require forcing the XCB platform
-   plugin for ``pyvistaqt`` to work correctly. You can set
-   ``QT_QPA_PLATFORM=xcb`` directly in your script before importing
-   Qt-related modules:
-
-   .. code:: python
-
-      import os
-
-      os.environ["QT_QPA_PLATFORM"] = "xcb"
-
-      import pyvista as pv
-      from pyvistaqt import BackgroundPlotter
-
 
 .. toctree::
    :maxdepth: 2
--- a/environment.yml
+++ b/environment.yml
@@ -10,6 +10,7 @@ dependencies:
   - pytest-cov
   - pytest-qt
   - qtpy>=1.9.0
+  - refleak>=0.2.1
   - scooby>=0.5.1
   - pip
   - pip:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,7 +95,7 @@ force-single-line = true
 [tool.ruff.lint.per-file-ignores]
 # Qt overrides and VTK-facing methods must use camelCase names, take fixed
 # (often unused) Qt event arguments, and pass raw GL booleans.
-"pyvistaqt/rwi.py" = ["N802", "N803", "ANN", "D102", "D105", "ARG002", "FBT003"]
+"pyvistaqt/rwi.py" = ["N802", "N803", "N815", "ANN", "D102", "D105", "ARG002", "FBT003"]
 "doc/**" = ["INP001"]
 "tests/**" = ["ANN001", "INP001", "S101", "T201", "PLR0915"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,10 +75,6 @@ line-length = 150
 
 [tool.ruff.format]
 docstring-code-format = true
-exclude = [
-  "pyvistaqt/rwi.py",  # should be maintained copy of upstream VTK code
-]
-
 [tool.ruff.lint]
 select = ["ALL"]
 ignore = [
@@ -87,9 +83,6 @@ ignore = [
   "D212",
   "ISC001",
   "PLR2004",
-]
-exclude = [
-  "pyvistaqt/rwi.py",
 ]
 
 [tool.ruff.lint.isort]
@@ -100,6 +93,9 @@ combine-as-imports = true
 force-single-line = true
 
 [tool.ruff.lint.per-file-ignores]
+# Qt overrides and VTK-facing methods must use camelCase names, take fixed
+# (often unused) Qt event arguments, and pass raw GL booleans.
+"pyvistaqt/rwi.py" = ["N802", "N803", "ANN", "D102", "D105", "ARG002", "FBT003"]
 "doc/**" = ["INP001"]
 "tests/**" = ["ANN001", "INP001", "S101", "T201", "PLR0915"]
 
@@ -138,7 +134,6 @@ branch = true
 source = ["pyvistaqt"]
 omit = [
     "*/setup.py",
-    "*/pyvistaqt/rwi.py",
 ]
 
 [tool.coverage.report]
@@ -155,9 +150,6 @@ python-version = "3.10"
 
 [tool.ty.src]
 include = ["pyvistaqt"]
-# rwi.py is a vendored copy of VTK's QVTKRenderWindowInteractor (it probes
-# every Qt binding), so it is intentionally not type-checked.
-exclude = ["pyvistaqt/rwi.py"]
 
 [tool.codespell]
 skip = "./.ci/*,./.git/*,./.hypothesis/*,./dist/*,./doc/examples/*,./doc/images/*,./docs/_build/*,./docs/images/*,./tests/tinypages/_build/*,.hypothesis*,*.doctree,*.eot,*.gif,*.html,*.inv,*.ipynb,*.jpg,*.js,*.json,*.mp4,*.mypy_cache/*,*.pickle,*.ply,*.png,*.pyc,*.ttf,*.txt,*.vti,*.vtk,*.vtu,*.woff,*.woff2,*.yml,*/_autosummary/*,*~,*cover,doc/_build/*,flycheck*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ test = [
     # Pull in whatever trame/nest_asyncio stack the installed pyvista needs
     # for VTKjs export. Each pyvista version pins its own compatible deps.
     "pyvista[jupyter]",
+    "refleak==0.2.1",
     "scooby==0.11.2",
     "sphinx_gallery==0.21.0",
 ]
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -670,6 +670,11 @@ class BackgroundPlotter(QtInteractor):
             LOG.debug("BackgroundPlotter.close app_window.close()")
             with contextlib.suppress(Exception):  # pragma: no cover
                 self.app_window.close()
+                # Schedule the window (and its child menus/toolbars/editor)
+                # for deletion. Otherwise those Qt objects outlive the close
+                # and their signal connections keep the closures they hold --
+                # which capture ``self`` -- alive, leaking the whole plotter.
+                self.app_window.deleteLater()
             LOG.debug("BackgroundPlotter.close done")
 
     def _close(self) -> None:
--- a/pyvistaqt/plotting.py
+++ b/pyvistaqt/plotting.py
@@ -10,9 +10,8 @@ Inheritance:
 
     BackgroundPlotter
     +-- QtInteractor
-        |-- QVTKRenderWindowInteractor
-        |   +-- QWidget  (the default QVTKRWIBaseClass; QGLWidget or
-        |                 QOpenGLWidget can be selected via vtkmodules.qt)
+        |-- QVTKRenderWindowInteractor  (pyvistaqt.rwi: FBO-based, no native
+        |   +-- QOpenGLWidget            window handle -- Wayland-compatible)
         +-- BasePlotter
 
     MainWindow
@@ -29,20 +28,20 @@ Implementation
     +-- QtInteractor.__init__(parent=self.frame)
         |-- with _no_base_plotter_init():
         |   +-- QVTKRenderWindowInteractor.__init__(parent=parent)
-        |       +-- QWidget.__init__(parent, wflags)
+        |       +-- QOpenGLWidget.__init__(parent)
         |-- BasePlotter.__init__(**kwargs)
         +-- self.ren_win: vtkRenderWindow = self.GetRenderWindow()
 
-Because ``QVTKRenderWindowInteractor.__init__`` calls ``QWidget.__init__``
-and the Qt bindings continue the cooperative ``super()`` chain past the Qt
-classes, this triggers a spurious ``BasePlotter.__init__`` call with no
-arguments. This cannot be solved by using ``super()`` because
-``QVTKRenderWindowInteractor.__init__`` does not use ``super()``, and the
-chaining happens inside Qt's ``QWidget`` initialization, outside of our
-control. We therefore temporarily monkey-patch ``BasePlotter.__init__`` with
-a no-op during the ``QVTKRenderWindowInteractor.__init__`` call (see
-``_no_base_plotter_init``), then call ``BasePlotter.__init__`` explicitly
-with the real arguments.
+Because ``QVTKRenderWindowInteractor.__init__`` calls
+``QOpenGLWidget.__init__`` and the Qt bindings continue the cooperative
+``super()`` chain past the Qt classes, this triggers a spurious
+``BasePlotter.__init__`` call with no arguments. This cannot be solved by
+using ``super()`` because ``QVTKRenderWindowInteractor.__init__`` does not
+use ``super()``, and the chaining happens inside Qt's widget initialization,
+outside of our control. We therefore temporarily monkey-patch
+``BasePlotter.__init__`` with a no-op during the
+``QVTKRenderWindowInteractor.__init__`` call (see ``_no_base_plotter_init``),
+then call ``BasePlotter.__init__`` explicitly with the real arguments.
 """  # noqa: D404
 
 from collections.abc import Callable
@@ -106,12 +105,6 @@ LOG = logging.getLogger("pyvistaqt")
 LOG.setLevel(logging.CRITICAL)
 LOG.addHandler(logging.StreamHandler())
 
-
-# for display bugs due to older intel integrated GPUs, setting
-# vtkmodules.qt.QVTKRWIBase = 'QGLWidget' could help. However, its use
-# is discouraged and does not work well on VTK9+, so let's not bother
-# changing it from the default 'QWidget'.
-# See https://github.com/pyvista/pyvista/pull/693
 
 # LOG is unused at the moment
 # LOG = logging.getLogger(__name__)  # noqa: ERA001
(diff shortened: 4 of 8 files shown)


larsoner:
@tkoyama010 do you want to try this? It's a big change but I think it's a sustainability win to use the same code paths as the VTK C++ code!

akaszynski:
@larsoner I’ll be checking this on one of my closed source apps and if it doesn’t have any issue I’ll approve and we can have this out on the next release. Sorry for the delay in review.

larsoner:
No problem! Would be great to have additional validation, thanks for looking when you get to it.

akaszynski:
Pushed three commits: the conflict resolution, a Wayland CI job, and one Qt 5 import fix. Details below, and please drop any of them you disagree with.

## 1. Merge conflict

Only `tests/test_plotting.py` conflicted, in five hunks — #854 had landed on main fixing the same macOS symptoms this branch does. Rather than picking a side I kept both protections in each hunk:

| hunk | resolution |
| --- | --- |
| `wait_exposed` | main's generator form + darwin pass on `QtBotTimeoutError`, this branch's `_no_gl()` guard and 30 s default (a superset of main's 10 s/5 s) |
| orbit thread join | this branch's `threading.Timer` skip + main's retry loop for the `threading._limbo` window, which the branch only suppressed once |
| `test_background_plotting_plots` | this branch's macOS software-GL skip + main's synchronous `ren_win.Render()`; took main's assertion, since main also changed the zoom the `0.9 < non_black < 1.0` bound was calibrated against and renamed the variable |

## 2. Wayland CI

Nothing here tested Wayland — every job runs xcb under Xvfb — so "It works with Wayland natively" and dropping the `QT_QPA_PLATFORM=xcb` docs caveat were untested claims. Added a job on both bindings against a headless Weston compositor. It went green on the first run:

```
Trying weston --backend=headless --renderer=gl
Compositor socket is up (weston --backend=headless --renderer=gl)
QGuiApplication.platformName() = 'wayland'
screen 'headless': 1920x1080
GL context: version 4.5, samples=0
```

`python pyvistaqt/rwi.py` then rendered the cone example and the suite started passing before I superseded the run with another push.

Two ways such a job can be green while testing nothing, both asserted up front in `.github/scripts/assert_wayland.py`:

- `QT_QPA_PLATFORM=wayland` is a request, not a guarantee — a missing plugin or a `wayland;xcb` fallback lands elsewhere silently, so it asserts `QGuiApplication.platformName()`.
- Weston's headless backend defaults to the pixman renderer, which advertises neither `wl_drm` nor `linux-dmabuf`; Mesa then hands out no EGL context and `QOpenGLWidget` has nothing to draw into. Hence `--renderer=gl`, plus an explicit `QOpenGLContext` check so a pixman fallback fails where the cause is legible.

## 3. `import pyvistaqt` fails on Qt 5

`from qtpy.QtOpenGLWidgets import QOpenGLWidget` raises on PyQt5/PySide2 — qtpy has no `QtOpenGLWidgets` there, it raises `QtBindingMissingModuleError` (an `ImportError`). main imports fine on Qt 5, so this is a regression, and no job here would catch it: the whole matrix is PyQt6/PySide6. `_get_event_pos` already has an explicit `except AttributeError:  # Qt5` branch, which is why I read this as an oversight rather than a deliberate drop — say the word if it was deliberate and I will drop the commit.

Qt 5 has had `QOpenGLWidget` since 5.4, just in `QtWidgets`, so the fallback is only about where the name lives. It is under `if TYPE_CHECKING:` / `else:` so the checker still sees one concrete base — resolved through a bare `try`/`except`, the name goes unresolved and two `ty: ignore` directives went unused, which is how I noticed.

Measured on PyQt5 5.15.14 / pyvista 0.48.4, `BackgroundPlotter` under `QT_QPA_PLATFORM=offscreen`: **main segfaults with no output; this branch raises `RenderWindowUnavailable` and exits cleanly.** Better behaved on Qt 5, not worse — it just could not be imported to find out.

## 4. Not fixed: `rw=` callers passing a platform render window

Reporting rather than pushing, since the fix is a design call.

`QVTKRenderWindowInteractor.__init__` still accepts `rw=`, then unconditionally calls `SetReadyForRendering(False)` and `SetFrameBlitModeToNoBlit()` on it. `SetReadyForRendering` is declared only on `vtkGenericOpenGLRenderWindow`:

```
vtkOpenGLRenderWindow has SetReadyForRendering:        False
vtkGenericOpenGLRenderWindow has it:                   True
```

So any caller passing `rw=vtkRenderWindow()` now gets an `AttributeError` from inside the constructor. That is not hypothetical — femorph-gui does exactly this to build an offscreen interactor, and it is on every platform, not just macOS, since the platform windows are all `vtkOpenGLRenderWindow` subclasses.

Given the widget is now FBO-backed, a plain platform render window arguably *cannot* work here, in which case a `TypeError` naming the requirement would beat an `AttributeError` on an internal call — but whether `rw=` should keep accepting anything at all is yours to decide.

## Suggestion

Given #3 and #4 both landed in Qt 5 / downstream territory that the matrix does not cover, a single PyQt5 job would be worth its cost — right now "Qt 5 still works" is an assumption, and it was already false on this branch.

.github/workflows/ci.yml
```diff
@@ -191,3 +191,126 @@ jobs:
...
+      # llvmpipe: the runners have no GPU, and the Wayland EGL path cannot fall
+      # back to OSMesa the way the Xvfb jobs do.
+      LIBGL_ALWAYS_SOFTWARE: "1"
+      GALLIUM_DRIVER: "llvmpipe"
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: "Install Weston and the Wayland/EGL runtime"
```
larsoner: I think we should port this over to setup-headless-display-action. Seems like it would be useful over there, and already has a `wm` option etc.

larsoner (commented):
Changes look reasonable to me, but I think the weston/wayland stuff more naturally fits in setup-headless-display-action. I'll move it and review more and push

larsoner:
@akaszynski good to go for you? Works in testing etc.?

akaszynski:
Not quite tested. This evening, sorry again about the delay.

akaszynski:
Correction: Claude tested it actually and it worked, but I'll interactively test to be certain.

larsoner:
It's okay no rush! I've had this open for weeks and it has sat idle :smile:

akaszynski:
Pushed 1f3d1e2, which fixes a crash this branch introduces. How we found it, since the path matters:

I was running [FEMORPH](https://www.femorph.com) (Qt5, Hyprland/Wayland) against this branch. The window came up and rendered fine, but opening a model died instantly:

```
Cannot make QOpenGLContext current in a different thread
Fatal Python error: Aborted
  rwi.py:322 in _cb_make_current
  pyvista/plotting/renderer.py in reset_camera
  ...
  femorph_gui/_threading.py:143 in protected_fn   <-- worker thread
```

FEMORPH loads meshes on a background thread, so `add_actor` reaches `reset_camera` -> `Render` -> `WindowMakeCurrentEvent` off the GUI thread. `QOpenGLContext` is thread-affine and Qt answers that with `qFatal`, so it's SIGABRT rather than an exception — nothing downstream can catch it.

Reduced to 15 lines with no FEMORPH involved:

```python
plotter = BackgroundPlotter()
threading.Thread(target=lambda: plotter.add_mesh(pyvista.Sphere())).start()
```

Then the control — same script, same pyvista/VTK/PyQt5, only pyvistaqt swapped:

| pyvistaqt | result |
| --- | --- |
| 0.11.4 | `SURVIVED`, exit 0 |
| this branch | `Cannot make QOpenGLContext current in a different thread`, exit 134 |

`git log -S'_cb_make_current'` puts it in 3005c6e, the first commit here. That's the mechanism: on `main` the widget hands VTK a native window handle (`SetWindowInfo(str(int(self.winId())))`) and VTK owns the context, so `MakeCurrent` is a plain GLX/EGL call with no thread affinity — `makeCurrent` doesn't appear in `main`'s `rwi.py` at all. Moving to `QOpenGLWidget` gives Qt the context and silently makes worker-thread rendering fatal.

The fix skips the event off-thread. It costs nothing: `currentContext()` is per-thread, so `IsCurrent()` already reports False there and VTK skips the GL work exactly as it does before the widget is realized; the GUI thread paints it on the next frame. Verified against FEMORPH — the mesh renders identically whether loaded on the worker or the GUI thread.

Also added `test_render_from_worker_thread`, because there was no coverage for rendering off the GUI thread (`threading` appears in the suite only to enumerate leaked threads). Worth having: without the guard it doesn't fail the test, it aborts the whole pytest run, so the guard reads like deletable dead code otherwise.

One note for anyone else testing on Wayland: PyQt5 wheels ship their own `libQt5WaylandClient`, and a distro `qtwayland` on `LD_LIBRARY_PATH` shadows it and fails to load with `undefined symbol: _ZTI27QPlatformServiceColorPicker` — Qt then falls back to xcb silently. Worth checking `app.platformName()` before concluding you've tested Wayland. Once actually on Wayland this branch looks great. :rocket:

akaszynski:
Two corrections to my comment above.

**The reproducer.** I said "15 lines" and then pasted a two-line sketch that isn't runnable. Here is the actual script:

```python
"""Render from a worker thread. Exits 0 on 0.11.4, aborts on this branch."""
import sys
import threading
import time

from qtpy.QtWidgets import QApplication

import pyvista as pv
import pyvistaqt
from pyvistaqt import MainWindow, QtInteractor

print("pyvistaqt:", pyvistaqt.__version__, flush=True)
app = QApplication(sys.argv)
win = MainWindow()
inter = QtInteractor(win)
win.setCentralWidget(inter)
win.resize(600, 400)
win.show()
for _ in range(60):
    app.processEvents()
    time.sleep(0.02)

err = []


def worker():
    try:
        inter.add_mesh(pv.Sphere(), reset_camera=True)
    except BaseException as exc:
        err.append(exc)


t = threading.Thread(target=worker)
t.start()
for _ in range(150):
    app.processEvents()
    time.sleep(0.02)
t.join(timeout=5)
print("SURVIVED; worker error:", err, flush=True)
```

```
$ python repro_thread.py            # pyvistaqt 0.11.4
pyvistaqt: 0.11.4
SURVIVED; worker error: []
$ echo $?
0

$ python repro_thread.py            # this branch
pyvistaqt: 0.13.0.dev37+g1c4d9fcaa
Cannot make QOpenGLContext current in a different thread
$ echo $?
134
```

**macOS.** 1f3d1e2 turned every macOS job red and 0870093 fixes that. The test, not the guard: VTK's render path segfaults when driven from a worker thread on macOS (`add_mesh` -> `add_actor` -> `reset_camera` -> `ResetCamera`, dying inside VTK on all four macOS runners while Linux and Windows pass). Skipped there, and invariant 6 now says so rather than implying the behavior is portable — the guard stops Qt aborting on a cross-thread `makeCurrent`, it does not make macOS render off the main thread.

akaszynski (approved):
Checked it locally and it worked great after the fixes. I have a high DPI laptop and on wayland it eliminated screen tearing on menu scrolling.

Please review my changes, but if you're good feel free to merge.

larsoner:
Okay I think it's worth merging this now. MNE-Python and some early `main` consumers can try it for a week or so, then I'll cut a release probably a week from Monday. Thanks for the iteration and testing @akaszynski !

PR #850, opened 2026-07-11
FIX: Remove VTK observer pins and drop the allow_bad_gc marker

larsoner:
Three independent leaks/failures kept tests on the VTK/plotter GC check's allow list; fix them and remove the marker entirely:

- QVTKRenderWindowInteractor bound-method observers: VTK holds a C++-side strong reference to each. Track the (subject, tag) pairs and remove them at teardown, including for never-realized widgets (show=False)
- trame server keeps a vtkWebApplication alive for the rest of the process by design, and its SynchronizationContext caches every exported data array behind a 20-second release window. Warm the server up in a session fixture so it pre-exists the GC snapshot, and clear the array cache before the check.
- MainWindow.close()/deleteLater() now tolerate an already-deleted C++ object: BackgroundPlotter.close schedules the window for deferred deletion, and pytest-qt's teardown then calls close() on the dead wrapper (RuntimeError under PySide and PyQt).
- test_ipython: del the plotter from the IPython user namespace; the InteractiveShell singleton otherwise pins it (and its VTK objects).

### DIFF
--- a/pyvistaqt/rwi.py
+++ b/pyvistaqt/rwi.py
@@ -288,10 +288,20 @@ class QVTKRenderWindowInteractor(QVTKRWIBaseClass):
         self._Timer = QTimer(self)
         self._Timer.timeout.connect(self.TimerEvent)
 
-        self._Iren.AddObserver('CreateTimerEvent', self.CreateTimer)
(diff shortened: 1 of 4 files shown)


larsoner:
This one is a straightforward cleanup so I'll mark for merge-when-green!

